Issue #5, opened 2017-07-11
Feature Request - Select CTP7 at command line for sca.py


bdorney:
Have sca.py take command line argument for which CTP7 it should connect to, right now it looks like it's hardcoded to eagle34.

jsturdy:
Have fix, needs cleanup and push

mexanick:
Implemented fix, issuing a PR

PR #7, opened 2017-08-17
Addressing issue #5 and clean up

mexanick:
- Some clean-up (rearranged lib files, left only the specific dependencies)
- Added .github with github templates and release tools
- Addressed issue https://github.com/cms-gem-daq-project/xhal/issues/5


.github/uploads.cfg
```diff
@@ -0,0 +1,7 @@
+{ "libwiscrpcsvc.so":"/home/mdalchen/work/xhal/lib/x86_64/libwiscrpcsvc.so",
+  "libxhal.so"      :"/home/mdalchen/work/xhal/lib/x86_64/libxhal.so",
+  "librwreg.so"     :"/home/mdalchen/work/xhal/lib/x86_64/librwreg.so",
+  "liblog4cplus.so" :"/home/mdalchen/work/xhal/lib/arm/liblog4cplus.so",
+  "liblmdb.so"      :"/home/mdalchen/work/xhal/lib/arm/liblmdb.so",
+  "libxerces-c.so"  :"/home/mdalchen/work/xhal/lib/arm/libxerces-c.so",
+  "libxhal_ctp7.so" :"/home/mdalchen/work/xhal/lib/arm/libxhal_ctp7.so"}
```
jsturdy: Is the absolute path here (as hard-coded) the desired behaviour?

.gitignore
```diff
@@ -7,5 +7,8 @@ Makefile~
 *.pyc
 python/*.pyc
 python/reg_interface/*.pyc
+python/daq_suite/*.pyc
+python/extras/*.pyc
+python/ctp7/*.pyc
 test/bin
 test/bin/*
```
jsturdy, line 14: Generally prefer a more "permit what's needed, ignore the rest" style, as easier to update

python/daq_suite/daq_suite/helper_main.py
```diff
@@ -1,5 +1,5 @@
 from rw_reg import *
-NOH=10
+NOH=1
```
bdorney, line 2: Is this representing the number of Optohybrids?

If so does having it hard coded to 1 limit us when working on test stands or in P5?

python/daq_suite/daq_suite/settings.py
```diff
@@ -25,7 +25,7 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = True
 
-ALLOWED_HOSTS = ['kvm-s3562-1-ip151-107.cms']
+ALLOWED_HOSTS = ['kvm-s3562-1-ip151-107.cms','cosmicstandtif.cern.ch']
```
bdorney: Should we also have gem904daq01 here?

python/reg_interface/sca.py
```diff
@@ -50,8 +50,8 @@ def main():
     ohMask = 0
     ohList = []
 
-    if len(sys.argv) < 3:
-        print('Usage: sca.py <oh_mask> <instructions>')
+    if len(sys.argv) < 4:
+        print('Usage: sca.py <card_name> <oh_mask> <instructions>')
```
bdorney, line 54: card_name here is the name of the CTP7? e.g. eagleXX?

bdorney (commented):
Minor questions requesting clarification

python/reg_interface/sca.py
mexanick: Yes, it is card's domain name. At P5 you can also eventually use the shelf-based name (although I don't really see why:))

python/daq_suite/daq_suite/settings.py
mexanick: Good point. Will add.

python/daq_suite/daq_suite/helper_main.py
mexanick: Well, this app is quite expert one and may clinch with our other software which uses IPBus, so the main usage for it would be NOT P5. And we currently have 1 board at the QC8 and integration stands, so to me this would be preferential.

.github/uploads.cfg
mexanick: Corrected. I thought it would be problematic to have env var there, but in fact it is not.

.gitignore
mexanick: I got your comment. While I will certainly address it in the future, for the moment I prefer to stay as it is. The structure of this project is quite complex and it will require significant time to address this issue.

mexanick:
All the comments are answered, new commit provided. Please take a look again @bdorney @jsturdy

python/daq_suite/daq_suite/settings.py
jsturdy: To this PR, or a future one?

mexanick: Ups, forgot. Will add now =)

jsturdy (approved):
Bon

PR #12, opened 2017-10-03
Adding test callers for V3 RPC calibration routines

mexanick:
## Description

Adding test callers for V3 RPC calibration routines
### Types of changes

## Motivation and Context

We need to test new RPC methods deployed with V3 electronics
## How Has This Been Tested?

Tested with Eagle60 and V3 electronics at QC lab
## Checklist:


python/reg_interface/test.py
```diff
@@ -42,32 +42,59 @@
...
+tscan.restype = c_uint
+tscan.argtypes = [c_uint, c_uint, c_uint, c_uint, c_uint, c_uint, c_uint]
+
 DEBUG = True
 
 def main():
   print "Start testing"
   start_time_ = timeit.default_timer()
-  rpc_connect("eagle34")
+  rpc_connect("eagle60")
```
jsturdy, line 63: This should just be made configurable now like the main `reg_interface` code, rather than leaving the hard-code in place indefinitely
If this script is specific for V3 (or a check on the firmware version isn't yet implemented) an additional flag/option could be considered for this purpose as well, though I strongly prefer that the software can figure out which version of the hardware is being communicated with and then acting accordingly.

jsturdy (changes requested):

python/reg_interface/test.py
mexanick: yeah, this is dirty test script, which I normally only use to quick check the various functionality. 
I will implement this later, along with big refactoring of the rpc_standalone library which is growing too big)

jsturdy (approved):

mexanick:
Closing as obsolete

Issue #16, opened 2017-10-06
Discussion: results array in genScan(...) and thresholdScan(...) is a local variable and may prove troublesome


bdorney:
## Brief summary of issue

I believe the current implementation of RPC methods `genScan(...)` and `thresholdScan(...)` will be more difficult to integrate with our existing repositories and use and cause unnecessary file I/O operations.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [X] Feature request (request for change which adds functionality)
- [X] Discussion

## Expected Behavior

The two scan applications [genScan(...)](https://github.com/cms-gem-daq-project/xhal/blob/447afc45fcac5670cd69bff9e82b350a321aaeb7/src/common/rpc_manager/calibration_routines.cc#L28) and [thresholdScan(...)](https://github.com/cms-gem-daq-project/xhal/blob/447afc45fcac5670cd69bff9e82b350a321aaeb7/src/common/rpc_standalone_client/rwreg.cpp#L93) should allow for a pointer of the `results` array to be supplied as an input argument rather than be a local variable.  

This should allow an array in python to be passed with the library is called and then the scan data is stored in the array.  Then integration with the current scan tools (and subsequent analysis) would be seamless, e.g.:

```
if v3El:
    genScan(scanData, ....)
else:
    scanData = getUltraScanResults(ohboard, options.gtx, THRESH_MAX - THRESH_MIN + 1, options.debug)
```

Then scanData can be parsed accordingly inside the ultra scan script and be neatly stored into the current output `TTree`.

I think this will offer a more seamless integration with the current `vfatqc-python-scripts`.   As shown in my original PR: https://github.com/cms-gem-daq-project/xhal/pull/13/files. From [scanDAC](https://github.com/bdorney/xhal/blob/0df591e8449c2f7d8c3fea08ee0a01ae26b11415/src/common/rpc_standalone_client/rwreg.cpp#L494) of my original PR:

```
DLLEXPORT uint32_t scanDAC(uint32_t* result, ssize_t size,
        unsigned int ohN, unsigned int vfatMask, unsigned int chan, std::string scanReg,
        unsigned int nEvts, unsigned int scanMin, unsigned int scanMax, unsigned int stepSize){
    req = wisc::RPCMsg("calibration_routines.genScan");
    req.set_word("ohN",ohN);
    req.set_word("mask",vfatMask);
    req.set_word("ch",chan);
    req.set_word("scanReg",scanReg);
[...]

PR #21, opened 2017-10-10
Addressing issue #16

mexanick:
## Description

Implementing flattened array return in genScan instead of text file creation
Addressing https://github.com/cms-gem-daq-project/xhal/issues/16

### Types of changes

## Motivation and Context

Required for seamless integration into the scan routines

## How Has This Been Tested?

It wasn't tested although I'm pretty sure from my previous experience that it will work. 
Some things to pay attention: 

- test.py shows an example how to use
- it is user responsibility to check if the array size will be an integer number (i.e. (scanMax-scanMin+1)/scanStep is an integer number

## Checklist:

### DIFF
--- a/include/xhal/rpc/calibration_routines.h
+++ b/include/xhal/rpc/calibration_routines.h
@@ -4,6 +4,6 @@
 #include "xhal/rpc/utils.h"
 
 DLLEXPORT uint32_t ttcGenConf(uint32_t L1Ainterval, uint32_t pulseDelay);
-DLLEXPORT uint32_t genScan(uint32_t nevts, uint32_t ohN, uint32_t dacMin, uint32_t dacMax, uint32_t dacStep, uint32_t ch, uint32_t enCal, uint32_t mask, char * scanReg);
+DLLEXPORT uint32_t genScan(uint32_t nevts, uint32_t ohN, uint32_t dacMin, uint32_t dacMax, uint32_t dacStep, uint32_t ch, uint32_t enCal, uint32_t mask, char * scanReg, uint32_t * result);
 
 #endif
--- a/python/reg_interface/test.py
+++ b/python/reg_interface/test.py
@@ -49,7 +49,7 @@ ttcGenConf.argtypes = [c_uint, c_uint]
 
 genScan = lib.genScan
 genScan.restype = c_uint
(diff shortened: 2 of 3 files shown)


bdorney:
> it is user responsibility to check if the array size will be an integer number (i.e. (scanMax-scanMin+1)/scanStep is an integer number

We should probably add a check that this integer requirement is enforced.  e.g. something like:

```
if( floor((scanMax-scanMin+1)/scanStep) != ceil((scanMax-scanMin+1)/scanStep)){
    cout<<"Improper usage case\n";
    cout<<"The quantity (scanMax-scanMin+1)/scanStep must be an integer\n";
    cout<<errorFlagUsage;
    return;
}
```

(Adapt above syntax as appropriate)

I'm not sure if this should be placed in the `genScan(...)` function in this PR or in `getUltraScanResults(...)` which will call `genScan(...)`.  In the later case the above syntax will be adapted to python.  I guess here is more appropriate because the usage case that needs to be checked is in `genScan(...)`.

bdorney (changes requested):
needs usage case check

mexanick:
@bdorney , setting correct range and step of the scan is caller's responsibility. So the code snippet you put above has to be adapted to python and reside in `getUltraScanResults(...)`

Now please clarify which changes you have requested. "needs usage case check" is not quite clear to me. Do you want a test case? You can run it with test.py - it is written there. I can do it for you when the V3 system is configured and ready to use.
"

bdorney:
The needs usage case check is the code I showed in my comment [above](https://github.com/cms-gem-daq-project/xhal/pull/21#issuecomment-335694942).

> @bdorney , setting correct range and step of the scan is caller's responsibility. So the code snippet you put above has to be adapted to python and reside in getUltraScanResults(...)

I agree that a caller should setup a function responsibly.   But a function being called *should* notify if it is being used improperly.  At the very least there should be some comment or documentation with the function that informs the user of this integer requirement.  Right now there is none in the code.

mexanick:
@bdorney in order to call the genScan method, you **must supply an array for resulting data**. How are you going to create an array of non-integer size???

bdorney:
> @bdorney in order to call the genScan method, you must supply an array for resulting data. How are you going to create an array of non-integer size???

Very carefully.

(Yeah fair point, I was not thinking).

bdorney (approved):

PR #26, opened 2017-10-13
New Feature: Interface for broadcastRead & Write

bdorney:
## Description

Added two new interfaces in xhal for optohybrid functionality:

- `broadcastRead(...)`
- `broadcastWrite(...)`

### Types of changes

## Motivation and Context

Allows for implementing broadcastRead/Write in OHv2b (FW bcast) or OHv3(SW done in `ctp7_modules`)

## How Has This Been Tested?

It hasn't yet.  I will test it once the corresponding broadcastRead(...) for `ctp7_modules` is written.

## Checklist:

### DIFF
--- a/include/xhal/rpc/optohybrid.h
+++ b/include/xhal/rpc/optohybrid.h
@@ -1,0 +1,9 @@
+#ifndef OPTOHYBRID_H
+#define OPTOHYBRID_H
+
+#include "xhal/rpc/utils.h"
+
+DLLEXPORT uint32_t broadcastRead(uint32_t ohN, char * regName, uint32_t vfatMask, uint32_t * result);
+DLLEXPORT uint32_t broadcastWrite(uint32_t ohN, char * regName, uint32_t value, uint32_t vfatMask);
+
+#endif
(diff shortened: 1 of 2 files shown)


mexanick (changes requested):
Please add details on interface specification: what you expect to be a register name? Is it a full name? or is it only a part of name related to VFAT registers?

bdorney:
> Please add details on interface specification: what you expect to be a register name? Is it a full name? or is it only a part of name related to VFAT registers?

I would like it to mirror the current uhal functionality.  The user should provide the VFAT Node name, e.g.:

- For OHv2b I would supply `VThreshold1`
- For OHv3 I would supply `CFG_THR_ARM_DAC` and *not* only a substring of this. 

While OHv3 does not have a broadcast it is assumed in:

https://github.com/cms-gem-daq-project/ctp7_modules/issues/9
https://github.com/cms-gem-daq-project/ctp7_modules/issues/10

That the "broadcastRead" in the case of OHv3 is just a loop over all non-masked VFATs.

For the "broadcastWrite" in the case of the OHv3 again it is just a loop over all non-masked VFATs but the order of the returned array should not be altered, e.g. if `VFAT7` is masked `result[7]` does not correspond to `VFAT8` but is just a dummy (e.g. 0 or -1) value for `VFAT7`.

mexanick (approved):

PR #30, opened 2017-11-06
 New Feature: ultra scan modules for v2b electronics

bdorney:
## Description

I have expanded the scan tools for v3 electronics to also work with v2b electronics. The modules identify at runtime what behavior should be executed.

Requires: https://github.com/cms-gem-daq-project/ctp7_modules/pull/13

### Types of changes

## Motivation and Context

If we are moving toward RPC/XHAL framework for CTP7 we should have the tools for both v2b and v3 electronics rather than having to move back and forth between libraries.

## How Has This Been Tested?

A complete working example is shown here:
http://cmsonline.cern.ch/cms-elog/1020178

For the v2b electronics I took a per channel threshold scan using the ipbus/uhal framework and then compared the results obtained when taking data with the rpc/xhal tools shown here. As can be seen the behavior is identical.

Additionally as shown the tools also work on v3 electronics.

### Screenshots (if appropriate):
See above elog post.

## Checklist:

- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,3 +2,17 @@
 XHAL interface library
 
 Developers documentation is generated with Doxygen and can be found under doc/
+
+# Compilation Instructions
+
+To compile xhal RPC modules execute:
+
+```
+make rpc
+```
+
+To compile CTP7 RPC libraries first compile `$BUILD_HOME/ctp7_modules` then execute:
+
+```
+make clean -f Makefile_ctp7
+```
--- a/src/common/rpc_manager/calibration_routines.cc
+++ b/src/common/rpc_manager/calibration_routines.cc
@@ -1,16 +1,37 @@
 #include "xhal/rpc/calibration_routines.h"
 
+DLLEXPORT uint32_t ttcGenToggle(uint32_t ohN, bool bRun){
+    wisc::RPCSvc* rpc_loc = getRPCptr();
+    req = wisc::RPCMsg("calibration_routines.ttcGenToggle");
+    req.set_word("ohN", ohN);
+    req.set_word("bRun", bRun);
+    try {
+        rsp = rpc_loc->call_method(req);
+    }
+    STANDARD_CATCH;
+
+    if (rsp.get_key_exists("error")) {
+        return 1;
+    }
+    return 0;
+} //End ttcGenToggle(...)
+
 /***
  * @brief configure TTC generator
  */
-DLLEXPORT uint32_t ttcGenConf(uint32_t L1Ainterval, uint32_t pulseDelay)
+DLLEXPORT uint32_t ttcGenConf(uint32_t ohN, uint32_t mode, uint32_t type, uint32_t pulseDelay, uint32_t L1Ainterval, uint32_t nPulses, bool bRun)
 {
     wisc::RPCSvc* rpc_loc = getRPCptr();
     //wisc::RPCMsg req_loc, rsp_loc;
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -2,3 +2,17 @@
...
+
+```
+make rpc
+```
+
+To compile CTP7 RPC libraries first compile `$BUILD_HOME/ctp7_modules` then execute:
+
+```
+make clean -f Makefile_ctp7
+```
```
mexanick, line 18: I have created Wiki in the repo and I think this should go there. Please consider such update

src/common/rpc_manager/calibration_routines.cc
```diff
@@ -52,30 +99,13 @@ DLLEXPORT uint32_t genScan(uint32_t nevts, uint32_t ohN, uint32_t dacMin, uint32
...
-    {
-        for (unsigned int j = 0; j<vfatSize; j++)
-        {
-            fprintf(f,"%d    %d    %d    %d\n",i,result[i*vfatSize+j]&0xFFFF,dacMin+j*dacStep,(result[i*vfatSize+j]>>16)&0xFFFF);
-        } 
-    }
-    */
+
     return 0;
 }
```
mexanick, line 111: I see you have removed the printing to file, but you didn't provide any other way to transfer the results. I suggest to change the return type to `uint32_t*` and return the `result` array.

mexanick (changes requested):

src/common/rpc_manager/calibration_routines.cc
bdorney: I'm not sure why github is flagging this as it is already like this in the develop branch, see [lines 63-79](https://github.com/cms-gem-daq-project/xhal/blob/develop/src/common/rpc_manager/calibration_routines.cc#L63).  I think this is because I forked your develop to get the `broadcastRead/Write()` behavior and so it's thinking that this is a new change (which it is not).

Moreover the input array is supplied as a pointer as we discussed originally and this is how the data is accessible to the user.  That's how the screenshots were generated in the above elog.

mexanick: ah, ok. I think you can safely remove the commented lines there (63-79)

bdorney: Will do!

bdorney: Ah that's what I did originally...they are already removed! :D

mexanick: Great! SO please move the Read.me update to the wiki part (it also follows md syntax) and I'll sign  off.

bdorney:
Actually I want to test with a latency scan first to make sure that no changes are needed when using an external trigger source

mexanick:
@bdorney, is this a final revision of the PR?

README.md
bdorney: I'll leave it in the `README.md` but I've added the info into a wiki page:
https://github.com/cms-gem-daq-project/xhal/wiki/Compilation-Instructions

src/common/rpc_manager/calibration_routines.cc
bdorney: Corresponding wiki page:
https://github.com/cms-gem-daq-project/xhal/wiki/Compilation-Instructions

mexanick (approved):

PR #35, opened 2018-01-23
Update setup.sh

jsturdy:
Make sourcing from outside the repo base stable

### Types of changes

- [x] Enhancement

## Motivation and Context
Currently, if the setup script is sourced from anywhere other than the project directory, things will *not* work as desired

## How Has This Been Tested?
Verified that sourcing from outside the project directory results in the `XHAL_ROOT` variable being set properly

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1,4 +1,3 @@
-export XHAL_ROOT=$PWD
+export XHAL_ROOT=$(dirname "${BASH_SOURCE[0]}")
 export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$XHAL_ROOT/lib
 export PATH=$PATH:$XHAL_ROOT/python/reg_interface
-# echo $XHAL_ROOT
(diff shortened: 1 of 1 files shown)


jsturdy:
Or maybe not... further testing needed

mexanick:
please resubmit PR to develop branch too. After this and the XML parser will be updated, things will go into master and new release.

jsturdy:
Can do, but as per the git-flow model, hotfixes are made to `master`

mexanick:
agree that hotfixes should go to `master`, but why this should be a hotfix? It is not even a bugfix, it is an enhacement...

jsturdy:
Hotfix part is that the env at p5 was not being set up

jsturdy:
Which leads to a secondary question, are current developments in `develop` breaking compatibility with current p5 operations? If so, a new release/maintenance branch for V2/P5 will need to be made for continued future compatibility

mexanick:
no, the current developments aren't breaking compatibility.

mexanick (approved):

PR #45, opened 2018-02-14
Propagating Andrew's improvements (address table pickling)

mexanick:
## Description

Implemented loading the address table from the pickled file if exist.
### Types of changes

## Motivation and Context

Pickling of the previously parsed address table significantly reduce the time required by reg_interface to start
## How Has This Been Tested?

Tests at eagle60 completed successfully
## Checklist:

### DIFF
--- a/python/reg_interface/rw_reg.py
+++ b/python/reg_interface/rw_reg.py
@@ -1,4 +1,3 @@
-import lxml.etree as xml
 import sys, os, subprocess, socket
 from time import sleep
 from ctypes import *
@@ -108,7 +107,7 @@ def main():
     print len(kids), kids.name
 
 def parseXML():
-    print 'Open pickled address table if available ',ADDRESS_TABLE_TOP,'...'
+    print 'Open pickled address table if available ',ADDRESS_TABLE_TOP[:-3],'pickle...'
 
     fname =  ADDRESS_TABLE_TOP[:-3] + "pickle"
     try:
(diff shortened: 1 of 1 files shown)


jsturdy (commented):
wasn't this one of the files that was migrating to the `reg-utils` repo?

mexanick:
It is. However the `reg-utils` repo is not ready for use yet and I plan to make a legacy release of xhal next week and want to have this improvement included. Then we will switch to the `reg-utils`. Apart from changes to the `reg-utils` repo we will have to make certain changes to `gem_ops` scripts as right now it takes this repo and its version of the reg_interface to rsync with the card, that's why I want to deliver a legacy version before switching to the new repo.

mexanick:
@jsturdy @bdorney @evka85 @andrewpeck 
I have tested the updated code at the eagle60. Curren version works well. Please find below some notes (they will be added to the documentation as well).
1. There's no lxml python module at the CTP7, so the pickle file has to be created at the host and copied to the CTP7. Pickle files are perfectly cross-platform, I've tested and no issues found.
2. I have updated the scripts at eagle60 to the most recent ones, which improves the speed and stability of the code (there were about 1 year old scripts there!). The new set of scripts uses GEM_PATH env variable instead of hardcoded path the library and xml/pickle files. It is pointing to the write-protected area by default. So if you need to play with different versions of lib and/or xml/pickle files you need to make a test area and point the GEM_PATH there.

jsturdy (commented):
What's responsible for creating the `pickle`d data here?

mexanick:
oh, I forgot to mention: `pickle`d file is created by `parseXML` method if it fails to locate the `gem_amc_top.pickle`. This method is called when one launches the `reg_interface.py` script.

jsturdy (approved):

PR #47, opened 2018-02-15
Adding update_lmdb method to reg_interface

mexanick:
## Description

Every time the XML address table receives update, the LMDB address table should be updated as well. Given PR adds a methos `update_lmdb` to the `reg_interface.py`

Usage:
after connecting to a given CTP7 card from the host PC, run:
```
update_lmdb /full/path/on/CTP7/to/new/address_table.xml
```
### Types of changes

## How Has This Been Tested?

Tested at `gem904qc8daq`

## Checklist:

### DIFF
--- a/python/reg_interface/reg_interface.py
+++ b/python/reg_interface/reg_interface.py
@@ -470,6 +470,14 @@ class Prompt(Cmd):
             print 'Incorrect usage.'
             return
 
+    def do_update_lmdb(self, args):
+        """Updates LMDB address table at the CTP7. USAGE: update_lmdb <absolute path name to the AMC xml address table at the CTP7>"""
+        if 'eagle' in hostname:
+            print 'This function can only be run from a host PC'
+        else:
+            update_atdb(args)
+            print 'LMDB address table updated'
+
     def execute(self, other_function, args):
         other_function = 'do_'+other_function
         call_func = getattr(Prompt,other_function)
--- a/python/reg_interface/rw_reg.py
+++ b/python/reg_interface/rw_reg.py
@@ -54,6 +54,10 @@ else:
   rList.restype = c_uint
   rList.argtypes=[POINTER(c_uint32),POINTER(c_uint32)]
 
+  update_atdb = lib.update_atdb
+  update_atdb.argtypes = [c_char_p]
+  update_atdb.restype = c_uint
+
   ADDRESS_TABLE_TOP = os.getenv("XHAL_ROOT")+'/etc/gem_amc_top.xml'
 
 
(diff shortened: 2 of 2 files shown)


python/reg_interface/reg_interface.py
```diff
@@ -470,6 +470,14 @@ def do_unmask(self, args):
             print 'Incorrect usage.'
             return
 
+    def do_update_lmdb(self, args):
+        """Updates LMDB address table at the CTP7. USAGE: update_lmdb <absolute path name to the AMC xml address table at the CTP7>"""
+        if 'eagle' in hostname:
```
jsturdy, line 475: Why not locally?

jsturdy (commented):
This leaves open the possibility that someone could set up the `LMDB` to be filled with a testing address table.
* What happens if two people attempt to update the DB quasi-simultaneously?
* How do you know which address table has been loaded into the `LMDB`?

As such, my question and comment from the previous PR#46 remain outstanding

python/reg_interface/reg_interface.py
mexanick: Because that functionality requires more changes in the code =)
In particular the librwreg.so should be changed. Eventually will be added later

mexanick:
> This leaves open the possibility that someone could set up the LMDB to be filled with a testing address table.

Yes. However this is an expert action and shouldn't be a problem.

>What happens if two people attempt to update the DB quasi-simultaneously?

There's a locking mechanism. So at the end there will be only one structurally correct address table in the DB, however I can't say for sure whether it will be the one from first user or from second.

>How do you know which address table has been loaded into the LMDB?

You don't... This is a good question and we should invent something. One of the easy solutions is to add the dummy register with no RW permissions and either address or mask field set to encode the version. Or we leave as it is due to the fact that the address table changes are rare. Or we can add a some kind of log.

> This should probably point to the "system" address table, no @mexanick?
Or request input from the user asking the path to the address table to associate with the LMDB

User must provide the full path to the new address table.

python/reg_interface/reg_interface.py
jsturdy: OK, makes sense :-)
Instead of that, I would still go my proposed route, of having the DB updated automatically, and if a non-standard address table is desired, some form of an on-demand DB/address table flag be provided in the tool (granted, this second part also *definitely* has some required code changes and *may* not even be viable)

mexanick: The automated update can be implemented in the ctp7user script (just call reg_interface in executable mode) whenever new FW release is coming. Then manual update can be used by experts for testing.

bdorney: Can you provide an example for this:

> The automated update can be implemented in the ctp7user script (just call reg_interface in executable mode) whenever new FW release is coming. Then manual update can be used by experts for testing.

mexanick: This will require a little change and I believe will deserve another PR which should be correlated to the ctp7user repo and eventually to the FW flashing script on the CTP7. The usage will be similar to:
```
reg -e update_atdb
```

PR #54, opened 2018-03-09
v3.1.0 Release preparation

mexanick:
Few changes in view of v3.1.0 release
## Description

**New features**:
* `rwc` method in `reg_interface.py`: read all registers containing keyword supporting wild card. *USAGE*: `rwc <KeyWord>`
* `rtest` method in `reg_interface.py`: read register certain number of times with a given interval in us. *USAGE*:` rtest <regName> <number of times> <interval>`
* `wtest` method in `reg_interface.py`: writes 1 to a register certain number of times with a given interval in us. *USAGE*: `rtest <regName> <number of times> <interval>`

**Breaking changes**:
* RPC manager library update according to name conventions described in [`ctp7_modules`](https://github.com/cms-gem-daq-project/ctp7_modules/issues/6). 

**Bugfixes**:
* Adjust uploads list for releasing
 
### Types of changes

## Motivation and Context

Provides some requested feature and makes code better maintanable
## Checklist:

### DIFF
--- a/python/reg_interface/reg_interface.py
+++ b/python/reg_interface/reg_interface.py
@@ -3,6 +3,8 @@ from cmd import Cmd
 import sys, os, subprocess
 from rw_reg import *
 from vfat_config import *
+import time
+usleep = lambda x: time.sleep(x/1000000.0)
 
 try:
     import readline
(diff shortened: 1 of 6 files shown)


jsturdy:
Would it make more sense to have the write test write random data?
Suppose it depends on what this test is supposed to accomplish/probe

mexanick:
Well, I wrote it quickly to have some watchdog stability tests... I don't think it is very useful indeed, but since I already wrote it, I don't see the reason to exclude it from the PR. The choice to write 1 was made because it was developed to test a kind of write-only regs where it doesn't matter what do you write

bdorney:
> The choice to write 1 was made because it was developed to test a kind of write-only regs where it doesn't matter what do you write

In some rare cases I am not sure that writing 1 will be a valid test, e.g. imaging the case where we have some signal inversion/flip in the firmware.  Will writing 1 be valid?  What about a register whose initial value is 1?

I would propose to add the ability to write a pattern of 1's and 0's on top of current functionality, the user could specify the pattern e.g. "pattern = 0b1010101"

And then it would write it and read it back.  If this is too involved it could be tabled to a future release/new feature PR.  I am not really sure on it's overall value.  Would probably need input from @evka85 or @andrewpeck.

python/reg_interface/reg_interface.py
```diff
@@ -319,17 +321,27 @@ def do_dumpKW(self, args):
...
-        """Read all registers containing KeyWord. USAGE: readKW <KeyWord>"""
-        if getNodesContaining(args) is not None and args!='':
-            for reg in getNodesContaining(args):
-                address = reg.real_address
-                if 'r' in str(reg.permission):
-                    print hex(address).rstrip('L'),reg.permission,'\t',tabPad(reg.name,7),readReg(reg)
-                elif reg.isModule: print hex(address).rstrip('L'),reg.permission,'\t',tabPad(reg.name,7) #,'Module!'
-                else: print hex(address).rstrip('L'),reg.permission,'\t',tabPad(reg.name,7) #,'No read permission!' 
-        else: print args,'not found!'
+    def do_rwc(self, args):
```
bdorney, line 324: For the user does this mean the `kw <X>` syntax is now to be changed to `rwc`? If so I would prefer to keep backwards compatibility to still allow:

- `readKW`
- `kw`

So that user's are not befuddled by the change (e.g. "change is transparent to the user").

bdorney (changes requested):
Require clarification regarding backwards compatibility of `reg_interface` keywords.

mexanick:
The `readKW` and `kw` methods remain intact.

On Mar 12, 2018 10:17, Brian Dorney <notifications@github.com> wrote:

@bdorney requested changes on this pull request.

Require clarification regarding backwards compatibility of reg_interface keywords.

bdorney (approved):

mexanick:
I don't think it requires attention in current release

On Mar 12, 2018 10:14, Brian Dorney <notifications@github.com> wrote:

The choice to write 1 was made because it was developed to test a kind of write-only regs where it doesn't matter what do you write

In some rare cases I am not sure that writing 1 will be a valid test, e.g. imaging the case where we have some signal inversion/flip in the firmware. Will writing 1 be valid? What about a register whose initial value is 1?

I would propose to add the ability to write a pattern of 1's and 0's on top of current functionality, the user could specify the pattern e.g. "pattern = 0b1010101"

And then it would write it and read it back. If this is too involved it could be tabled to a future release/new feature PR. I am not really sure on it's overall value. Would probably need input from @evka85<https://github.com/evka85> or @andrewpeck<https://github.com/andrewpeck>.

PR #29, opened 2017-11-06
New Feature: Keyword readAddress prints corresponding node name

bdorney:
## Description

The keyword `readAddress` now prints the corresponding node name in addition to the read back value.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Motivation and Context

Provides a debugging tool, if we see something like:
```
read memsvc error: Bus error accessing 0x65400208
```
In `/var/log/remote/eagleXX/messages.log` it would be helpful to know what node name this hex address corresponds to.

## How Has This Been Tested?

```
[dorney@cosmicstandtif]~/scratch0/CMS_GEM/CMS_GEM_DAQ/xhal/xhal/python/reg_interface% python reg_interface.py
Parsing /afs/cern.ch/user/d/dorney/scratch0/CMS_GEM/CMS_GEM_DAQ/xhal/xhal/etc/gem_amc_top.xml ...
connect eagle26Starting CTP7 Register Command Line Interface. Please connect to CTP7 using connect <hostname> command
CTP7 > connect eagle26
eagle26 > readAddress 0x654004b0
0x654004b0 	 node:  GEM_AMC.OH.OH0.GEB.VFATS.VFAT1.VFATChannels.ChanReg27
0x654004b0 	0xdeaddead
```
Note here `OH0` does not exist (which would why it would trigger a read bus error).

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #58, opened 2018-04-16
Update trigger OH main monitoring regs

mexanick:
Modyfying trigger OH main monitoring registers according to changes in PR #29 in ctp7_modules

### Types of changes

## Motivation and Context

Required for monitoring software to be even with ctp7_modules
## Checklist:

### DIFF
--- a/python/daq_suite/daq_suite/helper_main.py
+++ b/python/daq_suite/daq_suite/helper_main.py
@@ -82,12 +82,12 @@ def getTRIGGEROHmain():
   displaystring=[]
   namelist=[]
   values = []
-  res = (c_uint32 * (6*NOH))()
+  res = (c_uint32 * (8*NOH))()
   res_code = getRPCTRIGGEROHmain(res, NOH)
   if res_code == 0:
     values = [c for c in res]
   else:
-    for i in range(6*NOH):
+    for i in range(8*NOH):
       values.append(0)
 
   nextstr = ''
@@ -100,6 +100,8 @@ def getTRIGGEROHmain():
              'LINK1_MISSED_COMMA_CNT',
              'LINK0_OVERFLOW_CNT',
              'LINK1_OVERFLOW_CNT',
+             'LINK0_UNDERFLOW_CNT',
+             'LINK1_UNDERFLOW_CNT',
              'LINK0_SBIT_OVERFLOW_CNT',
              'LINK1_SBIT_OVERFLOW_CNT',]
   for i,regname in enumerate(namelist[1:]):
--- a/python/daq_suite/daq_suite/helper_main.py
+++ b/python/daq_suite/daq_suite/helper_main.py
@@ -82,12 +82,12 @@ def getTRIGGEROHmain():
   displaystring=[]
   namelist=[]
   values = []
-  res = (c_uint32 * (2*NOH))()
+  res = (c_uint32 * (6*NOH))()
   res_code = getRPCTRIGGEROHmain(res, NOH)
   if res_code == 0:
     values = [c for c in res]
   else:
-    for i in range(2*NOH):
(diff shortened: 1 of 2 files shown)


evka85:
Hi Misha,

If I understand it right, you're swapping out monitoring of LINK{0,1}_NOT_VALID_CNT for:
LINK{0,1}_MISSED_COMMA_CNT
LINK{0,1}_OVERFLOW_CNT
LINK{0,1}_SBIT_OVERFLOW_CNT

This is fine, but you could also add LINK{0,1}_UNDERFLOW_CNT.
As we talked before, these counters give different information. But the old LINK{0,1}_NOT_VALID_CNT is actually a subset of LINK{0,1}_MISSED_COMMA_CNT, and thus is redundant and has been removed.

Just to reiterate on the meaning:
  * LINK{0,1}_SBIT_OVERFLOW_CNT -- this is an interesting counter to monitor from operations perspective, but is not related to the health of the link itself. Rather it shows how many times OH had too many clusters which it couldn't fit into the 8 cluster per BX bandwidth. If this counter is going up it just means that OH is seeing a very high hit occupancy, which could be due to high radiation background, or thresholds configured too low.
  * The other 3 counters reflect the health of the optical links:
       -- LINK{0,1}_OVERFLOW_CNT and LINK{0,1}_UNDERFLOW_CNT going up indicate a clocking issue, specifically they go up when the frequency of the clock on the OH doesn't match the frequency on the CTP7. Given that CTP7 is providing the clock to OH, this in principle should not happen unless the OH is sending complete garbage and thus the clock cannot be recovered on CTP7 side, or the bit-error rate is insanely high, or the fiber is just disconnected, or OH is off. In other words, this indicates a critical problem. 
      -- LINK{0,1}_MISSED_COMMA_CNT also monitors the health of the link, but it's more sensitive, because it can go up due to isolated single bit errors. With radiation around, this might count one or two counts in a day or two. But if it starts running away, that would indicate a real problem, but in this case most likely the overflow and underflow counters would also see it.

I think it's important that the meaning is clear when these counters are reported to the user.
Also I think we should reset them during configure state transition (after TCDS has been configured), this can be done by writing to GEM_AMC.TRIGGER.CTRL.CNT_RESET.

Cheers,
Evaldas

mexanick:
missing registers are added

python/daq_suite/daq_suite/helper_main.py
```diff
@@ -82,12 +82,12 @@ def getTRIGGEROHmain():
   displaystring=[]
   namelist=[]
   values = []
-  res = (c_uint32 * (2*NOH))()
+  res = (c_uint32 * (8*NOH))()
```
bdorney, line 85: Should this be `12` and not `8`?

bdorney (commented):
small clarification needed

python/daq_suite/daq_suite/helper_main.py
mexanick: no, this is a number of registers to monitor for each link. It is NOH which can be 12

bdorney: Got it thanks

bdorney:
I guess a separate piece of code will be raising a warning or error state depending on the output of these register reads shown here?

Additionally the same would be true for @evka85's suggestion:

> Also I think we should reset them during configure state transition (after TCDS has been configured), this can be done by writing to GEM_AMC.TRIGGER.CTRL.CNT_RESET.

This PR covers only the monitoring of these registers not the follow up action once their values are known, correct?

mexanick:
Yes. The follow up action will be added later (perhaps to daq_suite, but possibly to another piece of SW)

bdorney (approved):
clarification provided.

PR #59, opened 2018-04-23
Reg interface prompt class separation

mexanick:
## Description

The command line prompt class defined previously in reg_interface.py had undergo substantial refactoring. According to request from @evka85 to have a generic, not specific to GEM reg_interface tool and a possibility to load on demand GEM-related functionality, the command line prompt class is now provided as a short version in ri_promt.py and as extended version (implemented as inherited class) in ri_prompt_extended.py 
The main reg_interface.py will search for the ri_prompt_extended.py and load it if found, otherwise will load the basic functionality only

### Types of changes

## Motivation and Context

CTP7 boards are used for different projects which would benefit of such tool, but don't need a GEM-specific methods
## How Has This Been Tested?

Modified tool was used to check both types of functionality using the gem904daqqc8 stand
## Checklist:


bdorney:
This is just a move of functionality from `reg_interface.py` into two different files?  Is that correct?  If so why?

mexanick:
because there're other users of CTP7 card who may benefit from the reg_interface while not needing the gem-related things. So gem-related methods have been decoupled.

bdorney:
> because there're other users of CTP7 card who may benefit from the reg_interface while not needing the gem-related things. So gem-related methods have been decoupled.

Looking at the deletions done to `reg_interface.py` it seems far more methods are removed than just gem-specific.  This seems a bit counter-intuitive based on the reasoning given.

From reviewing the deletions it seems only the following functions are GEM specific:

- `do_sbittranslate`
- `do_broadcastOH`
- `do_oh`
- `do_daq`
- `do_gemsystem`
- `do_trigger`
- `do_debug`
- `do_v2a`
- `do_mask`
- `do_unmask`
- `isValidOH`
- `isValidVFAT`

And then also `do_ttc` but this one seems to be only GEM specific because of the parent node name (e.g. `GEM_AMC`) but surely any other user would have similar/identical TTC block in the FW.  But yeah for now better to group this.

But I'm confused why other methods that don't appear to be GEM specific  (such as `do_connect `) are included in these deletions.  

Would suggest to keep in `reg_interface.py` to minimize churn or provide additional rational for their deletion.

bdorney (changes requested):
suggest that non-gem specific functions are not removed from `reg_interface.py` to minimize churn or provide additional reasoning behind their deletion.

mexanick:
**reg interface methods list**

|Method name | Generic | GEM-specific| Comment |
| :-------------: | :-------: | :--------------: | :----------: |
| `preloop` | Y | N | CMD prompt method |
| `postloop` | Y | N | CMD prompt method |
| `cmdloop_with_keyboard_interrupt `| Y | N | CMD prompt method |
| `do_exit` | Y | N | CMD prompt method |
| `execute` | Y | N | CMD prompt method |
| `complete_readGroup` | Y | N | tab completion |
| `complete_write` | Y | N | tab completion |
| `do_connect` | Y | N | generic RPC connection |
| `do_outputnode` | Y | N | show node description |
| `do_doc` | Y | N | alias for do_outputnode |
| `do_mpeek` | Y | N | low-level system address read |
| `do_mpoke` | Y | N | low-level system address write |
| `do_read` | Y | N | read a register |
| `do_readAddress` | Y | N | read modified address |
| `do_readRawAddress` | Y | N | read raw address |
| `do_readAll` | Y | N | read all registers with read permission |
| `do_readGroup` | Y | N | read all registers below node in reg tree |
| `do_kw` | Y | N | read all registers containing key word |
| `do_readKW` | Y | N | alias for do_kw |
| `do_rwc` | Y | N | read all registers containing wild card expression |
| `do_write` | Y | N | write a register |
| `do_dumpKW` | Y | N | dump addresses of all registers containing key word |
| `do_rtest ` | Y | N | test reading |
| `do_wtest ` | Y | N | test writing |
| `do_sbittranslate ` | N | Y | decode SBit cluster data |
| `do_broadcastOH ` | N | Y | broadcast write to multiple optical links |
| `do_oh ` | N | Y | execute certain command on a given OH |
| `do_daq ` | N | Y | read DAQ registers |
| `do_gemsystem ` | N | Y | read GEM_SYSTEM registers |
| `do_ttc` | N | Y | read TTC registers |
| `do_trigger` | N | Y | read trigger registers |
| `do_v2a` | N | Y | configure clocks for OH v2a |
| `do_mask` | N | Y | mask a VFAT(s) |
| `do_unmask` | N | Y | unmask a VFAT(s) |
| `do_update_lmdb` | N | Y | update LMDB for GEM |
| `do_debug` | N | Y | quick read of SBit clusters |
| `do_readFW` | N | Y | readFW versions |
| `do_fw` | N | Y | alias for do_readFW |
| `isValidOH ` | N | Y | helping method |
| `isValidVFAT ` | N | Y | helping method |

mexanick:
no method has been deleted, they were split between `ri_prompt.py` and `ri_prompt_extended.py` according to whether they are generic (then go to `ri_prompt.py`) or gem-specific

bdorney:
> no method has been deleted, they were split between ri_prompt.py and ri_prompt_extended.py according to whether they are generic (then go to ri_prompt.py) or gem-specific

This really clarifies things.  Thanks for this!

bdorney (approved):

PR #61, opened 2018-04-30
New Feature: Programming GBTs over fiber link

bdorney:
## Description

Adding @evka85's script for programming the GBTs over the fiber link.  Note a GBT must be minimally fused to lock to the fiber link for this to work and `GBT_READY` must be `0x1`.  Also the `I2C` jumper on the OH for the GBT of interest must have been removed.

### Types of changes

## Motivation and Context

Program GBTs over the fiber link.

## How Has This Been Tested?

Extensively during the v3 electronics testing before and after the ESR.

### DIFF
--- a/python/reg_interface/gbt.py
+++ b/python/reg_interface/gbt.py
@@ -1,0 +1,232 @@
+#!/usr/bin/env python
+
+from rw_reg import *
+from time import *
+import array
+import struct
+
+DEBUG=False
+
+class Colors:            
+    WHITE   = '\033[97m' 
+    CYAN    = '\033[96m' 
+    MAGENTA = '\033[95m' 
+    BLUE    = '\033[94m' 
+    YELLOW  = '\033[93m' 
+    GREEN   = '\033[92m' 
+    RED     = '\033[91m' 
+    ENDC    = '\033[0m'  
+
+ADDR_IC_ADDR = None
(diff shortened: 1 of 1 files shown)


mexanick:
do we want this functionality to be available only from CTP7, or from host PC as well?

mexanick (changes requested):
I think this PR will benefit from a little refactoring. In general, the code is very well written and the only down side is that the `main` function is supposed to be called from shell. Instead, we need to have options parser involved to maintain a standalone use case and rename `main` to smth like `programGBT` which will take all needed inputs as function arguments.

bdorney:
> do we want this functionality to be available only from CTP7, or from host PC as well?

Not really sure what @evka85's original idea was.  This was his script.  I'm including it so we can have it for the gemctp7user repo.

> I think this PR will benefit from a little refactoring. In general, the code is very well written and the only down side is that the main function is supposed to be called from shell. Instead, we need to have options parser involved to maintain a standalone use case and rename main to smth like programGBT which will take all needed inputs as function arguments.

I'll take a look to see how quickly that can be implemented.

bdorney:
`main()` has been changed to `programGBT(...)` with arguments.  The syntax of the command has remained unchanged.

mexanick (approved):

PR #68, opened 2018-05-11
Fix exe mode when executing on card

mexanick:
### Types of changes

## Motivation and Context

Fixing exe mode to work correctly on the card
## How Has This Been Tested?

tested at eagle61
examples of usage:
Examples from host:
```
reg_interface.py -e read "GEM_AMC.GEM_SYSTEM.BOARD_ID" -n eagle61
reg_interface.py -e write "GEM_AMC.GEM_SYSTEM.BOARD_ID 0xfeed" -n eagle61
```
Examples from card:
```
reg_interface.py -e read "GEM_AMC.GEM_SYSTEM.BOARD_ID"
reg_interface.py -e write "GEM_AMC.GEM_SYSTEM.BOARD_ID 0xfeed"
```
## Checklist:

### DIFF
--- a/python/reg_interface/reg_interface.py
+++ b/python/reg_interface/reg_interface.py
@@ -5,7 +5,7 @@ if __name__ == '__main__':
     from optparse import OptionParser
     parser = OptionParser()
     parser.add_option("-e", "--execute", type="str", dest="exe",
-                      help="Function to execute once", metavar="exe", default=None)
+                      help="Function to execute once, should be used as reg_interface -e <method> \"<method arguments>\" - note the quotes", metavar="exe", default=None)
     parser.add_option("-n", "--hostname", type="string", dest="hostname",
                       help="CTP7 hostname, default is the one used at p5", default="")
 
--- a/python/reg_interface/reg_interface.py
+++ b/python/reg_interface/reg_interface.py
@@ -7,7 +7,7 @@ if __name__ == '__main__':
     parser.add_option("-e", "--execute", type="str", dest="exe",
                       help="Function to execute once", metavar="exe", default=None)
     parser.add_option("-n", "--hostname", type="string", dest="hostname",
-                      help="CTP7 hostname, default is the one used at p5", default="amc-s2e01-23-03")
+                      help="CTP7 hostname, default is the one used at p5", default="")
 
     (options, args) = parser.parse_args()
 
@@ -20,9 +20,10 @@ if __name__ == '__main__':
 
     if options.exe:
         parseXML()
-        if (rpc_connect(options.hostname)):
-          print '[Connection error] RPC connection failed'
-          exit
+        if options.hostname:
+            if (rpc_connect(options.hostname)):
+                print '[Connection error] RPC connection failed'
+                exit
         prompt=Prompt()
         prompt.execute(options.exe,args)
         sys.exit()


python/reg_interface/reg_interface.py
```diff
@@ -7,7 +7,7 @@
     parser.add_option("-e", "--execute", type="str", dest="exe",
                       help="Function to execute once", metavar="exe", default=None)
     parser.add_option("-n", "--hostname", type="string", dest="hostname",
-                      help="CTP7 hostname, default is the one used at p5", default="amc-s2e01-23-03")
+                      help="CTP7 hostname, default is the one used at p5", default="")
```
jsturdy: is there a reason you have set `default=""` rather than just not having a `default`?

python/reg_interface/reg_interface.py
```diff
@@ -7,7 +7,7 @@
     parser.add_option("-e", "--execute", type="str", dest="exe",
                       help="Function to execute once", metavar="exe", default=None)
```
bdorney: In the usage example provided:

> reg_interface.py -e read "GEM_AMC.GEM_SYSTEM.BOARD_ID" -n eagle61
reg_interface.py -e write "GEM_AMC.GEM_SYSTEM.BOARD_ID 0xfeed" -n eagle61
Examples from card:
reg_interface.py -e read "GEM_AMC.GEM_SYSTEM.BOARD_ID"
reg_interface.py -e write "GEM_AMC.GEM_SYSTEM.BOARD_ID 0xfeed"

The keywords `read` and `write` are supplied with the `-e` option.  However the help menu does not illustrate that these keywords should be used with the `-e` option (I think this was my original problem when I was trying to use `reg_interface.py` outside of interactive mode).

Is there some way that either:

1. The help menu is updated to include a statement about the need for keywords, and what keywords are supported, and/or
2. At runtime the program checks if a keyword has been supplied with the `-e` option and if no keyword is supplied it throws an error (e.g. code `os.EX_USAGE`) along with a `print` statement illustrating what keywords are supported `I was expecting keywords from the set {'read','write', etc...}`.

bdorney (changes requested):
Add a way for the program to tell the user that the keywords from the set {`read`, `write`, etc...} must be used with the `-e` option.

python/reg_interface/reg_interface.py
mexanick: your problem was different. I was wrong usage of quotes for arguments of the method called interactively. It was not in using "-e" or "--execute"

python/reg_interface/reg_interface.py
mexanick: yes, there two reasons:
1). this executable mode should be available both at the card and at the host and the easy way to check is a check against whether the hostname has been provided, so the default value should be empty
2). a default doesn't really make sense as we are using the tool with various ctp7 cards

mexanick:
@bdorney I don't understand your comment. Any method can be called in exe mode...

python/reg_interface/reg_interface.py
bdorney: My problem is as described above.  The help menu did not illustrate effectively how the `-e` option should be used.  I tried:

```
reg_interface.py --hostname="eagle60" --execute="kw GEM_AMC.OH.OH0.FPGA.CONTROL.RELEASE.VERSION"
```

Which does not work.  And the help menu does *not* give enough information to deduce the syntax on the usage here:

```
% reg_interface.py -h
Usage: reg_interface.py [options]

Options:
  -h, --help            show this help message and exit
  -e exe, --execute=exe
                        Function to execute once
  -n HOSTNAME, --hostname=HOSTNAME
                        CTP7 hostname, default is the one used at p5
```

Nor does the program throw an exception which accurately explains the cause of the problem (bad syntax):

```
% reg_interface.py --hostname="eagle60" --execute "kw GEM_AMC.OH.OH0.FPGA.CONTROL.RELEASE.VERSION"
Extended prompt module found
Open pickled address table if available  Traceback (most recent call last):
  File 
xhal/python/reg_interface/reg_interface.py", line 27, in <module>
    prompt.execute(options.exe,args)
  File 
xhal/python/reg_interface/ri_prompt_extended.py", line 19, in execute
    call_func = getattr(Prompt,other_function)
AttributeError: class Prompt has no attribute 'do_kw GEM_AMC.OH.OH0.FPGA.CONTROL.RELEASE.VERSION'
```

From the help menu it is not clear that the keyword should be placed outside of quotation marks as shown in the usage example, not does the exception illustrate that either.

This should be addressed.

bdorney: Obviously from the usage example in this PR the correct syntax is:

```
% reg_interface.py --hostname="eagle60" --execute kw "GEM_AMC.OH.OH0.FPGA.CONTROL.RELEASE.VERSION"
Extended prompt module found
Open pickled address table if available xhal/etc/gem_amc_top.pickle...
0x6500000c None 	GEM_AMC.OH.OH0.FPGA.CONTROL.RELEASE.VERSION		
0x65000014 r 	GEM_AMC.OH.OH0.FPGA.CONTROL.RELEASE.VERSION.MAJOR	0x00000003
0x65000014 r 	GEM_AMC.OH.OH0.FPGA.CONTROL.RELEASE.VERSION.MINOR	0x00000001
0x65000014 r 	GEM_AMC.OH.OH0.FPGA.CONTROL.RELEASE.VERSION.BUILD	0x00000001
0x65000014 r 	GEM_AMC.OH.OH0.FPGA.CONTROL.RELEASE.VERSION.GENERATION	0x0000000b
```

bdorney:
> @bdorney I don't understand your comment. Any method can be called in exe mode...

The comment is that from the help menu of `reg_interface.py` the syntax expected when using the `-e` option is not illustrated.  Additionally if incorrect syntax is given the output text does not given information to adequately understand the root of the problem.

mexanick:
this is fixed in last commit, please see above

bdorney:
Missed the updated commit, excellent.  This is exactly what I was hoping for!

bdorney (approved):

python/reg_interface/reg_interface.py
bdorney: Could default it to `None` in this case.

Then local mode is understood if `hostname` is `None`.

As you wish.

jsturdy: If a `default` is not provided, the option when parsed **will** be `None`

Issue #57, opened 2018-04-13
Bug Report: Missing Requirements File for xhal?


bdorney:
## Brief summary of issue

On the P5 machine I've come across a few problems trying to use the v3 electronics.  The release instructions are not successful see updated comments in issue #55. 

I have been using the `cc7` virtual machine `gemvm-daqcc7` 

After compiling all libraries on `gem904qc8daq` (os: `cc7`) and placing them in:

```
$XHAL_ROOT/lib/arm
$XHAL_ROOT/lib/x86_64
```

Trying to run `reg_interface.py` seems to generate the following import errors:

```
ImportError: No module named lxml.etree
ImportError: No module named cmd
```

The `vm` mentioned and the `cc7` DAQ machines in 904 seem to use `python v2.7.5` while the daq machines in 904 seem to have these packages installed by default? 

### Types of issue

- [X] Bug report (report an issue with the code)
- [ ] Feature request (request for change which adds functionality)

## Expected Behavior

We should have a requirements file to assist with system installation.

## Current Behavior

No requirements file is present.

## Possible Solution (for bugs)

Add a requirements file.

## Context (for feature requests)

Takes awhile to setup the code or a new release.

## Your Environment

* Version used: b4980d841b4738740d5bc7c25bda52e537dec17d
* Shell used: /bin/bash

mexanick:
requirements would be handled in a future packaging release

bdorney:
Was this closed by #73 @mexanick ?

mexanick:
yes

PR #73, opened 2018-06-19
Adding requirements, packaging for arm, removing unnecessary setup and libs

mexanick:
Makefile for `xhalarm` is added, the packaging works
setup.sh and libs are removed. The x-compiled libs for xerces, lmdb and log4cplus are now under peta-stage 
## Description

Fixing #57 
### Types of changes

## Checklist:


python/reg_interface_gem/pkg/setup.cfg
```diff
@@ -50,4 +50,8 @@ exclude=build,lib,.tox,third,*.egg,docs,packages,.eggs
 
 [bdist_rpm]
 requires = python >= 2.6
+    reg_utils-rwreg
```
jsturdy: oh the pain! mixing `_` and `-` :-)

xhalarm/Makefile
```diff
@@ -0,0 +1,68 @@
+BUILD_HOME   := $(shell dirname `cd ../; pwd`)
+Project      := xhal
+Package      := xhal
+ShortPackage := xhal
```
jsturdy, line 4: what's the interplay when two packages have the same `ShortPackage` (i forget where this is all propagated to...)

jsturdy (commented):
Couple of quick comments

xhalarm/Makefile
mexanick: there's no problem here because this is the very same thing, just x-compiled for different archs

python/reg_interface_gem/pkg/setup.cfg
mexanick: not anymore =)) reg_utils is not there since latest update, only rwreg

mexanick:
Merge conflict resolved

Makefile
```diff
@@ -1,6 +1,7 @@
 SUBPACKAGES := \
         python \
-        xhalcore
+        xhalcore \
+	xhalarm
```
jsturdy, line 4: is there a whitespace difference here? `tab` vs. some number of `space`s?  i guess if `make` is parsing it properly then it's not a deal breaker

jsturdy (approved):

Makefile
mexanick: indeed! it was working, but I fixed it)

PR #74, opened 2018-06-21
Improve the version detection when making a release

jsturdy:
## Description
Updates the `gembuild` pointer to the latest version for extraction of build information based on the closest tag

### Types of changes

- [x] Enhancement (non-breaking change which adds functionality)

## Motivation and Context
Allows the releases to always be sequential in their machine understood versioning, such that a new RPM can be installed in the case that the version number hasn't changed (testing releases)

## How Has This Been Tested?
Installed at P5 on top of current RPMs

### DIFF
--- a/python/reg_interface_gem/pkg/setup.py
+++ b/python/reg_interface_gem/pkg/setup.py
@@ -21,10 +21,6 @@ def getreqs():
 def getVersion():
     __version__='___version___'
     __release__='___release___'
-    if __release__.split('.')[2]:
-        __prerel__='-{0:s}'.format(__release__.split('.')[2])
-    else:
-        __prerel__=''
     __buildtag__='___buildtag___'
     __gitrev__='___gitrev___'
     __gitver__='___gitver___'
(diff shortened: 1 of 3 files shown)


mexanick:
this is already propagated in parallel PR

python/reg_interface_gem/pkg/setup.py
```diff
@@ -21,18 +21,14 @@ def getreqs():
...
-        __prerel__=''
     __buildtag__='___buildtag___'
     __gitrev__='___gitrev___'
     __gitver__='___gitver___'
     __packager__='___packager___'
     __builddate__='___builddate___'
     with open("xhal/_version.py","w") as verfile:
         verfile.write("""
-## This file is generated automatically from xhal_reg_interface_gem setup.py
+## This file is generated automatically from cmsgemos_gempython setup.py
```
mexanick: noooo!

jsturdy: haha, oops, maybe you shouldn't have been so quick on the trigger!

mexanick: true) I wanna go home!!

mexanick (approved):

PR #80, opened 2018-07-06
Develop

mexanick:
Bugfixing lib loading which was preventing rpcman functions from having proper rpc connection
## Description

Load only one instance of libprcman in python loader code
### Types of changes

## How Has This Been Tested?

Tested with update_lmdb method on eagle64
## Checklist:


python/reg_interface_gem/core/reg_extra_ops.py
```diff
@@ -2,6 +2,15 @@
...
 lib = CDLL("librpcman.so")
+rReg = lib.getReg
+rReg.restype = c_uint
+rReg.argtypes=[c_uint]
+wReg = lib.putReg
+wReg.argtypes=[c_uint,c_uint]
+rpc_connect = lib.init
+rpc_connect.argtypes = [c_char_p]
+rpc_connect.restype = c_uint
+
```
bdorney, line 13: These should no longer be taken from `reg_utils.reg_interface.common.reg_base_ops` but imported from here?

bdorney (approved):

PR #83, opened 2018-08-03
Adding memhub

mexanick:
This PR adds code for the memhub library and provides neccessary changes to the ipbus code, removing the lock mechanism and introducing a proper semaphore
## Description

Notes about implementation:
- The new library and corresponding header have been copied to `$PETA_STAGE/lib/` and `$PET_STAGE/usr/include/` respectively. We should communicate in future that the library is added to the following official linux releases for GEMs, so it would be included under `$PETA_STAGE` by default. 
- According to the point above, `Makefile`s have been changed and includes and libraries are centrally sourced now
- Some (perhaps be not all) unnecessary includes are removed

### Types of changes

## Motivation and Context

Semaphore is required to avoid register access requests collisions at the ctp7 level
## How Has This Been Tested?

Tested at `eagle33` and `eagle34`
## Checklist:


xhalarm/Makefile
```diff
@@ -16,6 +16,7 @@ XHAL_VER_PATCH:=$(shell $(BUILD_HOME)/$(Project)/config/tag2rel.sh | awk '{split
 include $(BUILD_HOME)/$(Project)/config/mfCommonDefs.mk
 include $(BUILD_HOME)/$(Project)/config/mfZynq.mk
 include $(BUILD_HOME)/$(Project)/config/mfRPMRules.mk
+include $(BUILD_HOME)/$(Project)/config/mfZynq.mk
```
jsturdy: duplicate of L17?

mexanick: oups, indeed. Need to test the compilation so that compiler name and options are not overwritten. Should be fine without line 19 though

mexanick: checked and fixed. Thanks for spotting!

jsturdy (approved):

PR #88, opened 2018-08-23
Implementing class structure and boost::python wrappers

### LINKED ISSUE #86: Put on C++ rails to integrate properly with the cmsgemos project
## Brief summary of issue
Currently the library is written entirely in C-style with method exported to C-style DLLs via extern C. No classes or namespaces are used per se (however, the language is still C++ and certain parts are c++ code, like wisc::RPCSvc and wisc::RPCMsg). For the ease of use the library should be refactored to provide namespaces and classes. This can be implemented by writing a classes with all the current methods and provide `DLLEXPORT` wrappers around them for compatibility with python code, like
```
namespace a {
  class B {
  public:
    cpp_method(){printf("Hello World");}
  };
}
extern "C" c_method(a::B* b)
{
  b->cpp_method();
}
```

In this case it won't require major refactoring of python code using the xHAL. The other option (requiring more work) is to rewrite everything in regular C++ and use boost_python to provide python wrappers. However this will call upon changes in our python repositories as well. 

### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Your Environment

* Version used:
* Shell used:

mexanick:
## Description

- Providing base classes `xhal::XHALInterface` and `xhal::XHALDevice`
  - `XHALInterface` sets up logging and connects to remote board
  - `XHALDevice` in addition parses address table at the host PC and loads the `memory` and `extras` modules
- `VFAT3`, `OptoHybrid` and `AMC` are now inherited from `XHALDevice`
- `DaqMonitor`, `CalRoutines` and `Utils` are now inherited from `XHALInterface`
- A `boost::python` wrapper creating `xhalpy` module is provided

Features implemented in python wrapper:
- Generic class exposure for `XHALDevice` and `DaqMon` classes. Others to follow.
- `std::vector` to python `list` and `std::map` to python dictionary conversion
- Function overloading for functions taking different type of parameter
- Function overloading for functions taking different number of parameters following the same sequence or optional parameters

Test script is added to test the above functionalities.

Boost::python doc:
https://www.boost.org/doc/libs/1_64_0/libs/python/doc/html/tutorial/index.html
Types conversion:
https://stackoverflow.com/questions/6157409/stdvector-to-boostpythonlist

Project documentation will be updated sometimes next week
### Types of changes

## Motivation and Context

Fixes #86 
## How Has This Been Tested?

Tested with specially written test.

The project has been rewritten to comply with c++11 at max, so one should not use `devtoolset` for compilation, use the system compilers instead. 
The only required env at gem904daq is:
```bash
bash-4.2$ echo $LD_LIBRARY_PATH
/opt/xdaq/lib:/opt/wiscrpcsvc/lib:/home/mdalchen/work/cms-gem-daq-project/xhal/xhalcore/lib
```
in order to compile go under `xhal/xhalcore` and run `make _all`.
In order to run the tests either add the `xhal/xhalcore/lib` to python path, or copy the test.py there and run from there
## Checklist:


jsturdy:
Overall, the structure and usage of the classes is really not clear to me (and I think we will need to really crack it out before moving forward):
* How are devices and module interfaces related?
  * Shouldn't the module interfaces be members of a device?
  * Are the module interfaces to be used independently?
  * This way one would run a routine from the configuration module on a given device, and the software would semantically represent the hardware (which I think is something we should try to follow, at least as closely as we can)
* A device is also an interface?
* There's a ton of hierarchy that I think can be cleaned up by having a parent/child structure (we had though of doing this with the `uhal` devices too but never followed up..., so it might be beneficial as this is  probably going to be a rewrite of our current way of doing things
  * E.g., when doing an operation on a given `VFAT`, it should really already know which `Optohybrid` it is associated with
    * I suppose this is a limitation of the RPC modules themselves, being on the CTP7 and the class objects being on the host PC (though having the hierarchy as private member variables instantiated at construction would probably be the way to go here)
  * Additionally, doing things this way *might* make it possible to selectively parse the address table (of course with some work) to avoid unnecessary bits, at least where the address table is relevant and associate with the given device only nodes that are relevant
* There is additional cleanup that I'll avoid mentioning here, because it's "old" code, and not related to the refactoring you've just done

At the end of the day I think what I'm going to want is the similarity of what currently exist in the `uhal` GEM device classes in `cmsgemos`:
* SW object representations of HW devices that provide functions to HW functionality (`rpc` calls or `c++` routines)
  * We can have many very distinct, self contained RPC calls (a la the various functions currently in, e.g., `HwGenericAMC`) which can build into fuller functionalities, or many, probably very overlapping, RPC modules, to avoid the RPC overhead, but having significant code reuse, just because that's one of the limitations of using RPCs
* (possibly, though we do like using it heavily, so almost definitely) exported bindings of these same signatures to a `python` ecosystem

I think the primary usage of the current structure in anything other than a test stand is what I'm trying to suss out, as the current way of structuring seems geared towards the test bench type operation (probably bred out of the current status of get-things-done-now development) and we need to shift the focus to making the complete framework coherent and scalable

mexanick:
>Overall, the structure and usage of the classes is really not clear to me (and I think we will need to >really crack it out before moving forward):
>* How are devices and module interfaces related?

`XHALInterface` is a parent of everything, including `XHALDevice`
It provides following methods:
* `connect`: connect to the remote board by its domain name
* `loadModule`: load remote module in order to access its methods  
* `setLogLevel`: change the verbosity of logger

>  * Shouldn't the module interfaces be members of a device?

No, a device is an extension to interface, which has also an ability to perform barebone register access operations (both atomic and bulk) and which loads the specific to the device type remote modules upon construction

>  * Are the module interfaces to be used independently?

The `XHALInterface` is not supposed to be used alone, but its child classes like `DaqMonitor` or `CalRoutines` are to be used independently

>  * This way one would run a routine from the configuration module on a given device, and the software would semantically represent the hardware (which I think is something we should try to follow, at least as closely as we can)
> * A device is also an interface?

Yes

> * There's a ton of hierarchy that I think can be cleaned up by having a parent/child structure (we had though of doing this with the `uhal` devices too but never followed up..., so it might be beneficial as this is  probably going to be a rewrite of our current way of doing things
>  * E.g., when doing an operation on a given `VFAT`, it should really already know which `Optohybrid` it is associated with
>    * I suppose this is a limitation of the RPC modules themselves, being on the CTP7 and the class objects being on the host PC (though having the hierarchy as private member variables instantiated at construction would probably be the way to go here)

Current use cases suggests that we actually are not that much interested in a single VFAT operations, and our VFAT methods are smth like "configure all vfats"... However the thin layer under cmsgemos hardware managers could provide this if required

>  * Additionally, doing things this way *might* make it possible to selectively parse the address table (of course with some work) to avoid unnecessary bits, at least where the address table is relevant and associate with the given device only nodes that are relevant

Again, most of the operations are outsourced to the CTP7 and uses an LMDB address table which is parsed only once it is updated and later represents a key-value store with O(1) access time

> * There is additional cleanup that I'll avoid mentioning here, because it's "old" code, and not related to the refactoring you've just done

Please open a separate issue

> At the end of the day I think what I'm going to want is the similarity of what currently exist in the `uhal` GEM device classes in `cmsgemos`:
>* SW object representations of HW devices that provide functions to HW functionality (`rpc` calls or `c++` routines)
> * We can have many very distinct, self contained RPC calls (a la the various functions currently in, e.g., `HwGenericAMC`) which can build into fuller functionalities, or many, probably very overlapping, RPC modules, to avoid the RPC overhead, but having significant code reuse, just because that's one of the limitations of using RPCs

I'm not sure I understand which kind of limitations of RPC you mean... RPC doesn't provide any w.r.t. uhal...

> * (possibly, though we do like using it heavily, so almost definitely) exported bindings of these same signatures to a `python` ecosystem

The examples are already implemented and the wrapper will be of course extended to cover all the existing functionality. However there's a need to change certain interfaces. We can't use the `ctypes` in the way we did it before (however there's a possibility to use vectors and maps, converting them into python lists and dicts, which is a great improvement) and as this will require updates to python code I want to separate it from a current PR while providing all nuts abd bits for development here, so other developers could jump in if they fancy. 

> I think the primary usage of the current structure in anything other than a test stand is what I'm trying to suss out, as the current way of structuring seems geared towards the test bench type operation (probably bred out of the current status of get-things-done-now development) and we need to shift the focus to making the complete framework coherent and scalable 

I don't really see the limitations on scalability here. Again, a thin layer wrapper can (and probably should) be provided in XDAQ-based software to make it more semantic.

mexanick:
[xhal_inheritance.pdf](https://github.com/cms-gem-daq-project/xhal/files/2317603/xhal_inheritance.pdf)

jsturdy (approved):
Approved, based on outcome of discussion of future steps

PR #89, opened 2018-08-27
New Feature: expose bindings to python with boost::python

bdorney:
## Description

Exposes xhal classes to python.

Supporting changes:
- `XHALInterface` now has a public member function for returning the network alias of the AMC `getBoardName()`,
- `XHALInterface` and all inheriting classes will now store RPC modules that are to be loaded in `m_map_modName_modVer` 
- `XHALInterface` now has a public member function for (re-)loading all RPC modules which have been stored in `m_map_modName_modVer`.  Very useful for (re-)connecting.
- `XHALDevice` will now make use of inherited data member `m_map_modName_modVer` and inherited function `loadStoredModules` when (re-)connecting.
- New class `AMC` inheriting from `XHALDevice` defined in `amc.h` and `amc.cpp` in relevant locations.
  - Allows for getting VFAT mask for one or multiple links using corresponding CTP7 modules.
  - Allows for SBIT readout of a single OH using corresponding CTP7 module
- Added new utility structures for significantly reducing argument bloat in calibration scans:
   - `xhal::ParamScan` for providing scan parameters,
   - `xhal::ParamCalPulse` for providing calibration pulse parameters,
   - `xhal::ParamTtcGen` for providing configuration for the TTC Generator.
   - Defined in `../xhal/utils/CalParamTypes.h`
- Added new C++ types that will be exposed to python:
   - `PyDictUint32` is a typedef for `std::map<T, uint32_t>` where `T` is a template,
   - `PyDictVecUint32` is a typedef for `std::map<T, PyListUint32> where `T` is a template
       - Note:  `PyListUint32` is already defined and is of type `std::vector<uint32_t>`
   - `NestedPyDict` is a typedef for `std::map<T,M>` for T & M templates and M is typically itself an instance of `std::map<T,M>`, etc...
   - Defined in `../xhal/utils/PyTypes.h`
- Calibration routines now take as argument one or more Param structures as arguments.
- Calibration routines will no longer return an integer to be interpreted as an exit code but return sensible types (e.g. objects defined in `PyTypes.h` or void) and raise exceptions instead; additionally taking messy pointers to arrays as inputs has been dropped.
- `DaqMonitor` class has a new public method `getmonOHSCAmain` for returning SCA monitored temperatures and voltages for multiple OH links.
- Scan related methods of `Optohybrid` class will now take as input an instance of `xhal::ParamScan` to avoid argument bloat.
- Methods of `Optohybrid` class will no longer return an integer to be interpreted as an exit code but return sensible types (e.g. objects defined in `PyTypes.h` or void) and raise exceptions instead; additionally taking messy pointers to arrays as inputs has been dropped 
- Provided new methods to `VFAT3` class for DAC monitoring with the VFAT3 ADC.
- Methods of `VFAT3` class will no longer return an integer to be interpreted as an exit code but return sensible types (e.g. objects defined in `PyTypes.h` or void) and raise exceptions instead
- Added compiler preinstructions to be architecture safe when trying to expose `XHALXMLParser` to python.
 
Supercedes: https://github.com/cms-gem-daq-project/xhal/pull/87

### Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Motivation and Context

Exposing the C++ API to python to improve ease of use and reduce complexity.

Additionally now functions/class methods return sensible types and raise exceptions (breaking change).  To limit argument bloat scan related functions/class methods take struct's which specify a set of parameters.

## How Has This Been Tested?

Testing is being passed to several students in the lab and would like to get this merged in so they can compartmentalise tests and submit follow-up bug reports (PR patches).

## Checklist:

- [X] I have updated the documentation accordingly.

### DIFF
--- a/xhalcore/Makefile
+++ b/xhalcore/Makefile
@@ -87,7 +87,8 @@ $(OBJS_RPC_MAN):$(SRCS_RPC_MAN)
 	$(CC) $(CCFLAGS) $(ADDFLAGS) $(INC) $(LIB) -c $(@:%.o=%.cpp) -o $@ 
 
 $(XHALPY_LIB):$(OBJS_XHALPY)
-	$(CC) $(CCFLAGS) ${LDFLAGS} $(ADDFLAGS) $(INC) $(LIB) -lboost_python -L$(PYTHON_LIB_PREFIX) -l$(PYTHON_LIB) -Llib/ -lxhal -lrpcman -o $@ $<
+	#$(CC) $(CCFLAGS) ${LDFLAGS} $(ADDFLAGS) $(INC) $(LIB) -lboost_python -L$(PYTHON_LIB_PREFIX) -l$(PYTHON_LIB) -Llib/ -lxhal -lrpcman -o $@ $<
+	g++ $(CCFLAGS) ${LDFLAGS} $(ADDFLAGS) $(INC) $(LIB) -lboost_python -L$(PYTHON_LIB_PREFIX) -l$(PYTHON_LIB) -Llib/ -lxhal -lrpcman -o $@ $<
 
 $(OBJS_XHALPY):$(SRCS_XHALPY)
 	$(CC) $(CCFLAGS) $(ADDFLAGS) $(INC) -I$(PYTHON_INCLUDE_PREFIX) $(LIB) -c -o $@ $<
--- a/xhalcore/include/xhal/utils/CalParamTypes.h
+++ b/xhalcore/include/xhal/utils/CalParamTypes.h
@@ -1,0 +1,81 @@
+#ifndef CAL_UTILS_H
+#define CAL_UTILS_H
+
+namespace xhal {
+    struct ParamScan{
+        //Hardware selection
+        uint32_t ohN;
+        uint32_t ohMask;
+        uint32_t vfatN;
+        uint32_t vfatMask;
+        uint32_t chan;      //channel of interest
+
+        //Params
+        bool useUltra;
+        bool useExtTrig;
+
+        uint32_t dacMax;    //Maximum dac value
+        uint32_t dacMin;    //Minimum dac value
+        uint32_t dacSelect; //DAC to use for Monitoring
+        uint32_t dacStep;   //step size for dac
+        uint32_t nevts;     //Number of events
+        uint32_t waitTime;  //unit of time; uints depend on function
+
+        string scanReg;     //Register to scan against
+
+        ParamScan(){
+            useUltra = true;
+            useExtTrig = false;
+
+            dacMax=254;
+            dacMin=0;
+            dacStep=1;
+            nevts=100;
+        }
+    }; //End ParamScan
+
+    struct ParamCalPulse{
+        bool enable; //true (false) turn on (off) calpulse
+        bool isCurrent; //true (false) is current injection (voltage pulse)
+
+        uint32_t duration; //duration in BX's (CFG_CAL_DUR)
+        uint32_t extVoltStep; //External voltage step 0->disable; 1->enable (CFG_CAL_EXT)
+        uint32_t height; //height of calpulse (CFG_CAL_DAC)
+        uint32_t phase; //phase of calpulse (CFG_CAL_PHI)
+        uint32_t polarity; //polarity of calpulse 0->pos; 1->neg (CFG_CAL_SEL_POL)
+        uint32_t scaleFactor; //current pulse scale factor (CFG_CAL_FS)
+
+        SetupCalPulse(){
+            enable = false;
+            isCurrent = false;
+
+            duration = 0x1ff;
+            extVoltStep = 0x0;
+            height = 0x0;
+            phase = 0x0;
+            polarity = 0x0;
+            scaleFactor = 0x0;
+        }
+    }; //End ParamCalPulse
+
+    struct ParamTtcGen{
+        bool enable;
+
+        uint32_t L1Ainterval;
+        uint32_t mode;
+        uint32_t nPulses;
+        uint32_t pulseDelay;
+        uint32_t pulseRate;
+        uint32_t type;
+
+        ParamTtcGen(){
+            enable = false;
+
+            L1Ainterval = 250;
+            pulseDelay = 40;
+            pulseRate = 0;
+        }
+    }; //End ParamTtcGen
+
+} //End namespace xhal
+#endif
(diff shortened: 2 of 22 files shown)


jsturdy:
From my discussion with @mexanick 
* `xhal` library is:
  * `XHALInterface`: provide RPC connection and loading of RPC modules
  * `XHALDevice`: inherits from `XHALInterface`, provides `read`/`write`/`block read`/`block write` functionality over the RPC connection (but these will probably not be used much/ever, unless the `uhal` `IPBus` block transactions fail to deliver)
* `cmsgemos` codebase is:
  * `GEMHwDevice` (and specific descendant classes, e.g., `HwGenericAMC`, `HwGLIB`, `HwOptoHybrid`: contains an `XHALInterface` member object, defines all functional operations on the hardware, whether through single register requests via `uhal`, or through calls to RPC modules defined in the `ctp7_modules` and properly loaded through the `XHALInterface` member object, exposes these HW device objects and a subset of the functionality to `python` via `boost::python`
  * `<HW>Manager`: defines state transitions for hardware using the appropriate `GEMHwDevice`
  * `<HW>Monitor`: newly migrated into the monitoring framework @mexanick is currently working on

bdorney:
This description does *not* really set what direction should be taken with development, or how to go forward in the future.

Moreover, the current HEAD of develop bricks the v3 python API.  If instead of private discussions could we use the weekly ops meeting to discuss how to focus SW development.  Specifically it would be good to:

- Describe tasks that need to be done (providing pseudocode and examples when possible), and
- Allocate tasks to developers, and
- Define deadlines for PR's to be submitted.

Since I put a non-negligible amount of time into this PR and if it is not merged, or informed what follow-up actions should be taken, that time is wasted.

xhalcore/Makefile
```diff
@@ -87,7 +87,8 @@ $(OBJS_RPC_MAN):$(SRCS_RPC_MAN)
 	$(CC) $(CCFLAGS) $(ADDFLAGS) $(INC) $(LIB) -c $(@:%.o=%.cpp) -o $@ 
 
 $(XHALPY_LIB):$(OBJS_XHALPY)
-	$(CC) $(CCFLAGS) ${LDFLAGS} $(ADDFLAGS) $(INC) $(LIB) -lboost_python -L$(PYTHON_LIB_PREFIX) -l$(PYTHON_LIB) -Llib/ -lxhal -lrpcman -o $@ $<
+	#$(CC) $(CCFLAGS) ${LDFLAGS} $(ADDFLAGS) $(INC) $(LIB) -lboost_python -L$(PYTHON_LIB_PREFIX) -l$(PYTHON_LIB) -Llib/ -lxhal -lrpcman -o $@ $<
+	g++ $(CCFLAGS) ${LDFLAGS} $(ADDFLAGS) $(INC) $(LIB) -lboost_python -L$(PYTHON_LIB_PREFIX) -l$(PYTHON_LIB) -Llib/ -lxhal -lrpcman -o $@ $<
```
mexanick, line 91: What is the motivation of this change?

xhalcore/include/xhal/utils/CalParamTypes.h
```diff
@@ -0,0 +1,97 @@
+#ifndef CAL_UTILS_H
+#define CAL_UTILS_H
+
+#include <string>
+
+namespace xhal {
+    struct ParamCalPulse{
```
mexanick, line 7: The namespaces convention is broken here. While this file resides under xhal/utils, it should have extra nested namespace `utils`, e.g. `xhal::utils::ParamCalPulse`. 
I should add however, that IMHO this file should go under `xhal/rpc` as it only contains structs for RPC calls.

xhalcore/src/common/python_wrappers/XHALInterface_python.cpp
```diff
@@ -63,27 +81,147 @@ BOOST_PYTHON_MODULE(xhalpy){
   register_exception_translator<xhal::utils::XHALRPCException>(&translate_XHALRPCException);
   register_exception_translator<xhal::utils::XHALRPCNotConnectedException>(&translate_XHALRPCNotConnectedException);
 
+  /*class_<xhal::XHALInterface>("XHALInterface", init<const std::string&>())
```
mexanick, line 84: please provide a good reason why you have removed the XHALInterface bindings or revert this change

mexanick (changes requested):
Please perform a rebase so the latest developments I've pushed are included here. Also please address the other comments inline

xhalcore/src/common/python_wrappers/XHALInterface_python.cpp
bdorney: You told me on Mattermost to specifically not expose XHALInterface to python bindings.

If this decision has been reverted I can revert this change.

xhalcore/include/xhal/utils/CalParamTypes.h
bdorney: Sure, I'll add the namespace and move the file.

xhalcore/Makefile
bdorney: It did not compile with `gcc` but did with `g++` it was unclear as to why.  Specifically the python bindings did not compile.  The rest of xhal was fine.

xhalcore/src/common/python_wrappers/XHALInterface_python.cpp
mexanick: this was before I submitted a fix for the logging issue... Please reinclude it. An update to the the testing script is also desired to check whether things are still working.

bdorney:
@mexanick I can implement the changes you have proposed.  However @jsturdy has still listed this as "On Hold" and "Blocked."  So is that the case? i.e. implementing this changes at present seems like it will be a waste of time, or perhaps I've misunderstood @jsturdy.

xhalcore/Makefile
mexanick: It should compile just fine if you have the config submodule up to date. Please try to compile with clean environment on CC7. It should be enough to compile this package to use the default GCC.

bdorney: That's probably the issue then.  Which repo does the submodule live in?  I'm not aware when it gets updated so if you can point me to it I'll put it on my watch list so I get notifications.  Or maybe @jsturdy could use the Mattermost API to also watch it like he's done with the OHv3 FW releases.

mexanick:
I let @jsturdy comment. However I have nothing agains having this changes implemented and tested with python bindings, although we may or may not use all of them in future integration with `cmsgemos`.

jsturdy:
Closed, other route

PR #91, opened 2018-09-12
Adding XHALInterface constructor accepting external logger

mexanick:
### Types of changes

## Checklist:


jsturdy:
Is this waiting on anything?

xhalcore/Makefile
```diff
@@ -93,7 +93,7 @@ $(OBJS_XHALPY):$(SRCS_XHALPY)
 	$(CC) $(CCFLAGS) $(ADDFLAGS) $(INC) -I$(PYTHON_INCLUDE_PREFIX) $(LIB) -c -o $@ $<
 
 clean:
-	-${RM} ${XHALCORE_LIB} ${OBJS_UTILS} ${OBJS_XHAL} ${RPC_MAN_LIB} ${OBJS_RPC_MAN} ${OBJS_XHALPY}
+	-${RM} ${XHALCORE_LIB} ${XHALPY_LIB} ${OBJS_UTILS} ${OBJS_XHAL} ${RPC_MAN_LIB} ${OBJS_RPC_MAN} ${OBJS_XHALPY}
```
jsturdy: why `$(RM)` when `-rm` is used the next line?

mexanick:
on your, @jsturdy and @bdorney reviews =)

xhalcore/Makefile
mexanick: Historic reasons... Can go with any (and `${RM}` states for `rm -f` by default`) 
Let me know if you have strong preference

jsturdy: My preference is always consistency and uniformity when there isn't a reason for a difference :-)

xhalcore/include/xhal/XHALInterface.h
```diff
@@ -19,29 +19,28 @@
...
-#define WARN( MSG) LOG4CPLUS_WARN( m_logger, MSG)
-#define ERROR(MSG) LOG4CPLUS_ERROR(m_logger, MSG)
-#define FATAL(MSG) LOG4CPLUS_FATAL(m_logger, MSG)
-
+#define XHAL_TRACE(MSG) LOG4CPLUS_TRACE(m_logger, MSG)
+#define XHAL_DEBUG(MSG) LOG4CPLUS_DEBUG(m_logger, MSG)
+#define XHAL_INFO(MSG) LOG4CPLUS_INFO(m_logger, MSG)
+#define XHAL_WARN(MSG) LOG4CPLUS_WARN(m_logger, MSG)
+#define XHAL_ERROR(MSG) LOG4CPLUS_ERROR(m_logger, MSG)
+#define XHAL_FATAL(MSG) LOG4CPLUS_FATAL(m_logger, MSG)
```
bdorney, line 27: All of these errors below are catching on something from `wish::RPCSvc`.  Is there any reason why we are providing this wrapper?  e.g. why not call these all `RPC_TRACE`, ..., `RPC_WARN`, ..., etc...?

bdorney:
Since #89 has been blocked, and we have a reconnect method, can you incorporate `loadStoredModules()` from:

https://github.com/cms-gem-daq-project/xhal/pull/89/files#diff-a547ad5e911c04f37c34a25a514b49b5

To this PR?  And then have it called by both `connect` and `reconnect` as shown in the above PR.  This alleviated some issues I was having when using the exposed bindings in python.

xhalcore/include/xhal/utils/XHALXMLParser.h
```diff
@@ -37,12 +41,12 @@
...
-#define INFO( MSG) LOG4CPLUS_INFO( m_logger, MSG)
-#define WARN( MSG) LOG4CPLUS_WARN( m_logger, MSG)
-#define ERROR(MSG) LOG4CPLUS_ERROR(m_logger, MSG)
-#define FATAL(MSG) LOG4CPLUS_FATAL(m_logger, MSG)
+#define XHAL_TRACE(MSG) LOG4CPLUS_TRACE(m_logger, MSG)
+#define XHAL_DEBUG(MSG) LOG4CPLUS_DEBUG(m_logger, MSG)
+#define XHAL_INFO(MSG) LOG4CPLUS_INFO(m_logger, MSG)
+#define XHAL_WARN(MSG) LOG4CPLUS_WARN(m_logger, MSG)
+#define XHAL_ERROR(MSG) LOG4CPLUS_ERROR(m_logger, MSG)
+#define XHAL_FATAL(MSG) LOG4CPLUS_FATAL(m_logger, MSG)
```
bdorney, line 49: Ignorant question, but why do these have to be defined in two different locations instead of just being loaded from a header file in both cases?  E.g. they are defined here in `XHALXMLParser.h` and also above in `XHALInterface.h`

xhalcore/src/common/XHALInterface.cpp
```diff
@@ -15,46 +15,76 @@ xhal::XHALInterface::XHALInterface(const std::string& board_domain_name):
...
+  try {
+    this->connect();
+    XHAL_INFO("XHAL Interface connected");
+  }
+  catch (xhal::utils::XHALRPCException &e) {
+    XHAL_INFO("XHAL Interface failed to connect");
+  }
+}
+
+xhal::XHALInterface::XHALInterface(const std::string& board_domain_name, log4cplus::Logger& logger):
```
bdorney, line 29: Maybe I missed a commit somewhere, but wasn't the constructor here at least loading the baseline rpc modules? e.g. `memory.so`?

What is our plan for loading modules, does this happen in the cmsgemos class or here?

bdorney (changes requested):
Some clarifications needed.

xhalcore/include/xhal/XHALInterface.h
mexanick: there could be errors not associated with the RPC itself, which also has to be catched and transmitted correctly

mexanick:
> Since #89 has been blocked, and we have a reconnect method, can you incorporate `loadStoredModules()` from:
> 
> https://github.com/cms-gem-daq-project/xhal/pull/89/files#diff-a547ad5e911c04f37c34a25a514b49b5
> 
> To this PR? And then have it called by both `connect` and `reconnect` as shown in the above PR. This alleviated some issues I was having when using the exposed bindings in python.

No. It is not related to the scope of current PR.

xhalcore/src/common/XHALInterface.cpp
mexanick: No, it provides an interface and is a base class. Loading modules should be (and is) implemented in descendant classes, e.g. XHALDevice

xhalcore/include/xhal/utils/XHALXMLParser.h
mexanick: Will be refactored in future. There're more deep issues with logging, which may require more work

xhalcore/include/xhal/XHALInterface.h
bdorney: For sure I don't disagree with that statement, but it seems the implementation of these below is based solely on `wisc::RPCSvc::*Exception` objects.  In my mind an `xhal` exception/error/statement should be xhal specific and the below should be `RPC` specific e.g.:

 ```C++
#define RPC_TRACE(MSG) LOG4CPLUS_TRACE(m_logger, MSG)
#define RPC_DEBUG(MSG) LOG4CPLUS_DEBUG(m_logger, MSG)
#define  RPC_INFO(MSG) LOG4CPLUS_INFO(m_logger, MSG)
#define RPC_WARN(MSG) LOG4CPLUS_WARN(m_logger, MSG)
#define RPC_ERROR(MSG) LOG4CPLUS_ERROR(m_logger, MSG)
#define RPC_FATAL(MSG) LOG4CPLUS_FATAL(m_logger, MSG)

#define XHAL_TRACE(MSG) LOG4CPLUS_TRACE(m_logger, MSG)
#define XHAL_DEBUG(MSG) LOG4CPLUS_DEBUG(m_logger, MSG)
#define XHAL_INFO(MSG) LOG4CPLUS_INFO(m_logger, MSG)
#define XHAL_WARN(MSG) LOG4CPLUS_WARN(m_logger, MSG)
#define XHAL_ERROR(MSG) LOG4CPLUS_ERROR(m_logger, MSG)
#define XHAL_FATAL(MSG) LOG4CPLUS_FATAL(m_logger, MSG)
```

Where first set is the below and the second is XHAL specific...or maybe it's my ignorance and I have misunderstood the use case/intention.

xhalcore/include/xhal/utils/XHALXMLParser.h
bdorney: Okay.

xhalcore/src/common/XHALInterface.cpp
bdorney: Okay.

bdorney:
> No. It is not related to the scope of current PR.

Fair enough but I think it would be good to have this in a future PR.

xhalcore/include/xhal/XHALInterface.h
mexanick: I'm afraid you are mixing the logging with exception handling. The exceptions translation is done differently and reflects an origin of it. This, instead, is only logging macros.

bdorney (approved):

PR #93, opened 2018-09-28
New Feature: Added legacy xhal routine for DAC scans of VFAT3 (PreBoost)

bdorney:
## Description

HAL for DAC scans on v3 electronics.

### Types of changes

## Motivation and Context

See: https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/203

## How Has This Been Tested?

See: https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/203

## Checklist:

- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [X] I have added tests to cover my changes.

### DIFF
--- a/xhalcore/include/xhal/rpc/calibration_routines.h
+++ b/xhalcore/include/xhal/rpc/calibration_routines.h
@@ -42,7 +42,7 @@ struct vfat3DACSize{
 DLLEXPORT uint32_t checkSbitMappingWithCalPulse(uint32_t ohN, uint32_t vfatN, uint32_t mask, bool useCalPulse, bool currentPulse, uint32_t calScaleFactor, uint32_t nevts, uint32_t L1Ainterval, uint32_t pulseDelay, uint32_t *data);
 DLLEXPORT uint32_t checkSbitRateWithCalPulse(uint32_t ohN, uint32_t vfatN, uint32_t mask, bool useCalPulse, bool currentPulse, uint32_t calScaleFactor, uint32_t waitTime, uint32_t pulseRate, uint32_t pulseDelay, uint32_t *outDataCTP7Rate, uint32_t *outDataFPGAClusterCntRate, uint32_t *outDataVFATSBits);
 DLLEXPORT uint32_t dacScan(uint32_t ohN, uint32_t dacSelect, uint32_t dacStep, uint32_t mask, bool useExtRefADC, uint32_t * results);
-DLLEXPORT uint32_t dacScanMultiLink(uint32_t ohMask, uint32_t NOH=12, uint32_t dacSelect, uint32_t dacStep, bool useExtRefADC, uint32_t * results);
+DLLEXPORT uint32_t dacScanMultiLink(uint32_t ohMask, uint32_t NOH, uint32_t dacSelect, uint32_t dacStep, bool useExtRefADC, uint32_t * results);
 DLLEXPORT uint32_t genScan(uint32_t nevts, uint32_t ohN, uint32_t dacMin, uint32_t dacMax, uint32_t dacStep,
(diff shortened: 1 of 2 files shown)


xhalcore/include/xhal/rpc/calibration_routines.h
```diff
@@ -3,8 +3,46 @@
 
 #include "xhal/rpc/utils.h"
 
+struct vfat3DACSize{
+    uint32_t max[41];
+
+    vfat3DACSize(){
+        //ADC Measures Current
+        //I wonder if this dacMin and dacMax info could be added to the LMDB...?
```
mexanick, line 11: I'm against adding this info in an address table representation in LMDB because I want to keep it as light as possible. However, if there's a strong concern, we may add second LMDB which will keep only these values...

xhalcore/src/common/rpc_manager/calibration_routines.cc
```diff
@@ -94,6 +94,74 @@ DLLEXPORT uint32_t checkSbitRateWithCalPulse(uint32_t ohN, uint32_t vfatN, uint3
...
+    }
+    STANDARD_CATCH;
+
+    if (rsp.get_key_exists("error")) {
+        printf("Caught an error: %s\n", (rsp.get_string("error")).c_str());
+        return 1;
+    }
+
+    vfat3DACSize dacSize;
+    const uint32_t size = NOH * (dacSize.max[dacSelect] - 0+1)*24/dacStep;
```
mexanick: I don't understand this `dacSize.max[dacSelect] - 0+1`... typo?

bdorney: it was a placeholder to remind myself that the size should be:

24 * (max - min + 1) / step

Where min was always 0.

mexanick (changes requested):
the placeholder "-0 +1" must be removed

xhalcore/include/xhal/rpc/calibration_routines.h
mexanick: all this hardcoded info will make complicated future maintenance. Please consider other ways of retrieving it

bdorney: I thought the goal was that we wouldn't be maintaining this in the future; isn't this branch being depreciated after the updates to the HwDevice in cmsgemos?  Anyway that's secondary.

The key point would be that this is designed to solve a problem that won't exist in the future, namely we are forced to work with pointers to c-arrays with fixed size because the python API can't access this info otherwise.

In the future this method would be in cmsgemos, be returning from the `RPCMsg` a vector instead of writing to an array pointer.  And then that would be exposed to a python list or some other dynamic sized container.

In short `struct vfat3DACSize` won't exist in the future.  This will then be purely handled by the scan application.

bdorney: And final note, unless we modify the address table (either xml or lmdb) to include the max size of a register I'm not sure what else could be done.  Is that what is being requested?  

This should still be 100% solved once we can work with dynamic containers.

bdorney:
> the placeholder "-0 +1" must be removed

The `-0` I can remove, but the `+1` is required to get the index correct. :D

mexanick (dismissed):

bdorney:
@mexanick @jsturdy @BenjaminRS I had to add one last commit (ed9a23e) per VFAT3 team's instructions.  We should not scan `CFG_IREF` with the VFAT3's internal ADC's since this can only be scanned with an external ADC.

Additionally the VFAT3 Team mentioned that writing a value to `CFG_IREF` below `0x20` or above `0x32` may cause damage to the ASIC.

Review needed again.

mexanick (approved):

jsturdy (approved):

BenjaminRS (approved):

PR #99, opened 2018-11-29
[WIP] Implement RPC methods for configuring the GBTX's

lpetre-ulb:
## Description

This is the `xHAL`, branch `legacyPreBoost`, part of the [corresponding PR ](https://github.com/cms-gem-daq-project/ctp7_modules/pull/75)in `ctp7_modules` of which it depends.

See the parent PR for more informations.

### Types of changes

## Checklist:

### DIFF
--- a/xhalcore/include/xhal/rpc/gbt.h
+++ b/xhalcore/include/xhal/rpc/gbt.h
@@ -1,0 +1,11 @@
+#ifndef GBT_H
+#define GBT_H
+
+#include "xhal/rpc/utils.h"
+
+DLLEXPORT uint32_t scanGBTPhases(uint32_t ohN, uint32_t N, uint32_t phaseMin, uint32_t phaseMax, uint32_t phaseStep, uint32_t *results);
+DLLEXPORT uint32_t writeGBTConfig(uint32_t ohN, uint32_t gbtN, uint32_t configSize, uint8_t *config);
+DLLEXPORT uint32_t writeGBTPhase(uint32_t ohN, uint32_t vfatN, uint8_t phase);
+
+#endif
+
--- a/xhalcore/src/common/rpc_manager/gbt.cc
+++ b/xhalcore/src/common/rpc_manager/gbt.cc
@@ -1,0 +1,83 @@
+#include "xhal/rpc/gbt.h"
+
+DLLEXPORT uint32_t scanGBTPhases(uint32_t ohN, uint32_t N, uint32_t phaseMin, uint32_t phaseMax, uint32_t phaseStep, uint32_t *results){
+    req = wisc::RPCMsg("gbt.scanGBTPhases");
+
+    req.set_word("ohN", ohN);
+    req.set_word("N", N);
+    req.set_word("phaseMin", phaseMin);
+    req.set_word("phaseMax", phaseMax);
+    req.set_word("phaseStep", phaseStep);
+
+    wisc::RPCSvc* rpc_loc = getRPCptr();
+
+    try {
+        rsp = rpc_loc->call_method(req);
+    }
+    STANDARD_CATCH;
+
+    if (rsp.get_key_exists("error")) {
+        printf("Caught an error: %s\n", (rsp.get_string("error")).c_str());
+        return 1;
+    }
+
+    for (unsigned int vfatN = 0; vfatN < 24; vfatN++)
+    {
+        const std::string resultsKey = "OH" + std::to_string(ohN) + ".VFAT" + std::to_string(vfatN);
+        if (rsp.get_key_exists(resultsKey)){
+            ASSERT(rsp.get_word_array_size(resultsKey) == 16);
+            rsp.get_word_array(resultsKey, results + 16*vfatN);
+        }
+        else{
+            printf("Key '%s' not found.", resultsKey.c_str());
+            return 1;
+        }
+    }
+
+    return 0;
+} //End scanGBTPhase(...)
+
+DLLEXPORT uint32_t writeGBTConfig(uint32_t ohN, uint32_t gbtN, uint32_t configSize, uint8_t *config){
+    req = wisc::RPCMsg("gbt.writeGBTConfig");
+
+    req.set_word("ohN", ohN);
+    req.set_word("gbtN", gbtN);
+    req.set_binarydata("config", config, configSize);
+
+    wisc::RPCSvc* rpc_loc = getRPCptr();
+
+    try {
+        rsp = rpc_loc->call_method(req);
+    }
+    STANDARD_CATCH;
+
+    if (rsp.get_key_exists("error")) {
+        printf("Caught an error: %s\n", (rsp.get_string("error")).c_str());
+        return 1;
+    }
+
+    return 0;
+} //End writeGBTConfig(...)
+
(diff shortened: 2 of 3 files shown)


bdorney:
Until https://github.com/cms-gem-daq-project/ctp7_modules/pull/73 is merged in this should be considered a WIP.

bdorney:
@lpetre-ulb after some discussion with @mexanick and @jsturdy let's resurrect this PR.  This is necessary for connectivity testing pre-QC8 during super chamber construction.

xhalcore/include/xhal/rpc/gbt.h
```diff
@@ -0,0 +1,11 @@
+#ifndef GBT_H
+#define GBT_H
+
+#include "xhal/rpc/utils.h"
+
+DLLEXPORT uint32_t scanGBTPhases(uint32_t ohN, uint32_t N, uint32_t phaseMin, uint32_t phaseMax, uint32_t phaseStep, uint32_t *results);
```
bdorney, line 6: Single letter variable names should be avoided at all cost; change `N` to `nScans`.

xhalcore/src/common/rpc_manager/gbt.cc
```diff
@@ -0,0 +1,83 @@
+#include "xhal/rpc/gbt.h"
+
+DLLEXPORT uint32_t scanGBTPhases(uint32_t ohN, uint32_t N, uint32_t phaseMin, uint32_t phaseMax, uint32_t phaseStep, uint32_t *results){
+    req = wisc::RPCMsg("gbt.scanGBTPhases");
+
+    req.set_word("ohN", ohN);
+    req.set_word("N", N);
```
bdorney, line 7: I know this slipped into `ctp7_modules` but single letter variable names should be avoided at all cost.

xhalcore/include/xhal/rpc/gbt.h
bdorney: This could benefit from default parameter assignment, e.g.

```C++
DLLEXPORT uint32_t scanGBTPhases(uint32_t *results, uint32_t ohN, uint32_t N=100, uint32_t phaseMin=0, uint32_t phaseMax=15, uint32_t phaseStep=1);
```

xhalcore/include/xhal/rpc/gbt.h
```diff
@@ -0,0 +1,11 @@
...
+DLLEXPORT uint32_t writeGBTConfig(uint32_t ohN, uint32_t gbtN, uint32_t configSize, uint8_t *config);
```
bdorney, line 7: Why have `configSize` at all here? This size is always expected to be the same right?  It must be ~~356~~ 366?  This is a feature of the GBT.

xhalcore/src/common/rpc_manager/gbt.cc
```diff
@@ -0,0 +1,83 @@
...
+
+    try {
+        rsp = rpc_loc->call_method(req);
+    }
+    STANDARD_CATCH;
+
+    if (rsp.get_key_exists("error")) {
+        printf("Caught an error: %s\n", (rsp.get_string("error")).c_str());
+        return 1;
+    }
```
bdorney, line 22: This needs to be rethought since by scanning the VFAT phase you're going to introduce bad phases, and thus bad communication.  This will introduce `error` into the RPC response:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/78bfc744a7ba1cdb9c3d320bcbf5585c38553ffc/src/utils.cpp#L271-L275

So this will always be true.  Probably some thought in `ctp7_modules` is required.  I'm not sure if an RPC key can be removed once inserted (@mexanick)?

xhalcore/src/common/rpc_manager/gbt.cc
```diff
@@ -0,0 +1,83 @@
...
+
+    return 0;
+} //End scanGBTPhase(...)
+
+DLLEXPORT uint32_t writeGBTConfig(uint32_t ohN, uint32_t gbtN, uint32_t configSize, uint8_t *config){
+    req = wisc::RPCMsg("gbt.writeGBTConfig");
+
+    req.set_word("ohN", ohN);
+    req.set_word("gbtN", gbtN);
+    req.set_binarydata("config", config, configSize);
```
bdorney, line 45: `configSize` should probably just be hard coded.  With some appropriate check on `sizeof` the array `config` points to.

xhalcore/include/xhal/rpc/gbt.h
lpetre-ulb: Yes, I agree without you, that single letter variable was a bad idea...

xhalcore/include/xhal/rpc/gbt.h
lpetre-ulb: Right, the size is expected to always be 366. If it was pure C++, I would have used something like ` typedef std::array<uint8_t, CONFIG_SIZE> config_t;`, but since it is intended for CDLL Python wrapper, I felt more comfortable passing it real size (which is normally issued from `len(...)`) in order to avoid segmentation errors as much as possible. With a proper `boost::python` wrapper, your solution is obviously the best, but without, I feel it is more bug prone. Also note, that an incorrect config size triggers an error in the RPC module so it is easily noticed.

xhalcore/src/common/rpc_manager/gbt.cc
lpetre-ulb: Yes. Do you prefer that I leave it as is or that I correct it in the `ctp7_modules` ?

xhalcore/src/common/rpc_manager/gbt.cc
lpetre-ulb: Fortunately (or unfortunately ?), the `set_string` line is commented out so that the `error` key is not set. I have not catch this potential problem since the method currently works and I am not sure how to properly fix the (potential) issue.

xhalcore/src/common/rpc_manager/gbt.cc
lpetre-ulb: It would be best, but `sizeof` of `config` would just give the size of the pointer...

xhalcore/src/common/rpc_manager/gbt.cc
bdorney: Can you hotfix it in both the `master` and `develop` branches?

xhalcore/include/xhal/rpc/gbt.h
bdorney: Okay, I understand your rationale.  This will be fixed when it's migrated to `cmsgemos` anyway.

xhalcore/src/common/rpc_manager/gbt.cc
bdorney: I think this depends on the outcome of the discussion in https://github.com/cms-gem-daq-project/ctp7_modules/issues/84 & https://github.com/cms-gem-daq-project/ctp7_modules/issues/85.

For now let's comment these four lines out and open an issue here (in `xhal`) so we don't lose track.

xhalcore/src/common/rpc_manager/gbt.cc
bdorney: I should probably know this but what does `sizeof(*config)` give?

bdorney:
Superceded by #102

Issue #97, opened 2018-11-19
Package ARM libs with `devel` package`


jsturdy:
## Brief summary of issue
In order to build e.g., `ctp7_modules`, the `arm` libraries from `xhal` are required to be linked against.
The best way I can think of to accomplish this, would be to add them to the spec template for the `xhal-devel` package.  Then `xhal-devel` is a build time requirement of `ctp7_modules`

### Types of issue
- [x] Feature request (request for change which adds functionality)

## Expected Behavior
Would be best to not have to have 2-3 external packages required as source just to build.

## Current Behavior
The build process expects that `xhal` is checked out in the same working tree as, e.g., `ctp7_modules`

jsturdy:
See, e.g., a proposed library location in https://github.com/bdorney/ctp7_modules/blob/a539e76b8a5de26a29986fda3d027369f3c162d5/Makefile

jsturdy:
Closed by #100

PR #100, opened 2019-01-09
Cleanup build, package ARM libs in devel RPM

jsturdy:
## Description
This PR adjusts the build process to bundle the `arm` `libxhal.so` into the `devel` RPM, placing it in
`%{_prefix}/lib/arm/`

### Types of changes

## Motivation and Context
Some libraries, e.g., `ctp7_modules` require the `arm` version of `libxhal` to be available in order to link properly.  Addressing #97 

## How Has This Been Tested?
Updated libraries have been built (see cms-gem-daq-project/ctp7_modules#77)

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,12 +3,16 @@ SUBPACKAGES := \
 	xhalarm \
         xhalcore
 
-SUBPACKAGES.DEBUG    := $(patsubst %,%.debug,    ${SUBPACKAGES})
-SUBPACKAGES.RPM      := $(patsubst %,%.rpm,      ${SUBPACKAGES})
-SUBPACKAGES.DOC      := $(patsubst %,%.doc,      ${SUBPACKAGES})
-SUBPACKAGES.CLEANRPM := $(patsubst %,%.cleanrpm, ${SUBPACKAGES})
-SUBPACKAGES.CLEANDOC := $(patsubst %,%.cleandoc,    ${SUBPACKAGES})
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -39,3 +45,11 @@ $(SUBPACKAGES.CLEANDOC):
...
 $(SUBPACKAGES.CLEAN):
 	$(MAKE) -C $(patsubst %.clean,%, $@) clean
+
+.PHONY: $(SUBPACKAGES) $(SUBPACKAGES.INSTALL) $(SUBPACKAGES.CLEAN) $(SUBPACKAGES.DOC) $(SUBPACKAGES.RPM) $(SUBPACKAGES.CLEANRPM) $(SUBPACKAGES.CLEANDOC)
+
+python:
+
+xhalarm:
+
+xhalcore: xhalarm
```
mexanick, line 55: Why `xhalcore` depends on `xhalarm`? Also, isn't all these targets included in `make $(SUBPACKAGES)`?

mexanick (approved):
Conditionally approved. Clarification on one question requested

Makefile
jsturdy: It was done for the RPM building (might not be strictly necessary), since the `xhalcore` `xhal-devel` RPM needs the `xhalarm` `lib/libxhal.so`.
Enforcing it here just ensures that the library is built before `xhalcore` is built

jsturdy:
Closes #97

PR #108, opened 2019-04-10
update the packaging to make it compatible with ARM arch

mexanick:
Implementing changes to packaging required for ARM
## Description

* Adding `packageinfo.h`
* Modyfing `Makefile`s appropriately
* Adding postinstall script for ARM
* Bumping the config submodule

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #109, opened 2019-04-11
update the packaging to make it compatible with ARM arch

mexanick:
Identical to #108 
### Types of changes

### DIFF
--- a/xhalarm/include/packageinfo.h
+++ b/xhalarm/include/packageinfo.h
@@ -1,0 +1,11 @@
+#ifndef XHAL_PACKAGEINFO_H
+#define XHAL_PACKAGEINFO_H
+
+#ifndef DOXYGEN_IGNORE_THIS
+
+#define XHAL_REQUIRED_PACKAGE_LIST reedmuller
+#define XHAL_BUILD_REQUIRED_PACKAGE_LIST reedmuller-devel
+
+#endif
+
+#endif
--- a/xhalcore/include/packageinfo.h
+++ b/xhalcore/include/packageinfo.h
@@ -1,0 +1,11 @@
+#ifndef XHAL_PACKAGEINFO_H
+#define XHAL_PACKAGEINFO_H
+
+#ifndef DOXYGEN_IGNORE_THIS
+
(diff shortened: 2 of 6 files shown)


xhalarm/include/packageinfo.h
```diff
@@ -0,0 +1,11 @@
+#ifndef XHAL_PACKAGEINFO_H
+#define XHAL_PACKAGEINFO_H
+
+#ifndef DOXYGEN_IGNORE_THIS
+
+#define XHAL_REQUIRED_PACKAGE_LIST reedmuller
+#define XHAL_BUILD_REQUIRED_PACKAGE_LIST reedmuller-devel
```
jsturdy, line 7: Why is `reedmuller` required here (in `xhal`)?

xhalcore/include/packageinfo.h
```diff
@@ -0,0 +1,11 @@
+#ifndef XHAL_PACKAGEINFO_H
+#define XHAL_PACKAGEINFO_H
+
+#ifndef DOXYGEN_IGNORE_THIS
+
+#define XHAL_REQUIRED_PACKAGE_LIST reedmuller, xerces-c
```
jsturdy, line 6: Why is `reedmuller` required here?

xhalarm/include/packageinfo.h
mexanick: Ups, an oversight. However on the other hand, something has to be defined here otherwise the spec file will have an empty `Requires` tag. I think we can leave it as is for the moment, but add later the proper dependencies (lmdb, xerces-c) when we will have them packaged and installed with `smart`

xhalcore/include/packageinfo.h
mexanick: same as above. The dependencies can't be empty, but has to be updated in future. In this particular case (none-arm) I can remove it if you insist, but still the dependencies has to be updated. I think it deserves a different PR... but it is up to you

jsturdy (approved):
Hmm, OK, well this was why I hadn't included `include/packageinfo.h` in any PR yet :-P

Issue #111, opened 2019-04-23
Bug Report: Writing phase scan results to file only preserves the last unmasked OH.


bdorney:
## Brief summary of issue

There's a bug when writing the output phase results.

### Types of issue

- [x] Bug report (report an issue with the code)
- [ ] Feature request (request for change which adds functionality)

## Expected Behavior

The output file for subsequent OH's in `ohMask` should not overwrite the previous one.

## Current Behavior

https://github.com/cms-gem-daq-project/xhal/blob/3015276210f258f6da2f0ae03a1d5795abe9b317/python/reg_interface_gem/core/gbt_utils_extended.py#L42-L80

https://github.com/cms-gem-daq-project/xhal/blob/3015276210f258f6da2f0ae03a1d5795abe9b317/python/reg_interface_gem/core/gbt_utils_extended.py#L114-L135

If `gbtPhaseScan` is provided with a non-`None` `outputFile` then `saveGBTPhaseScanResults` will overwrite the phase scans for all OH's but the last one in `ohMask`. 

### Steps to Reproduce (for bugs)

1. Call `saveGBTPhaseScanResults` with an ohMask with at least 2 unmasked bits

## Possible Solution (for bugs)

Either `saveGBTPhaseScanResults` opens the output file in read/write mode (file is appended) rather than write (in which case the file is overwritten); or one file is made per unmasked OH.

## Context (for feature requests)

Trying to track phase scan results but can only see the last one...

## Your Environment

* Version used: 3015276210f258f6da2f0ae03a1d5795abe9b317
* Shell used:

bdorney:
Closed by #113

PR #113, opened 2019-04-27
Allow multiple GBT phase scan results to be written to the same file

lpetre-ulb:
## Description

This PR fixes https://github.com/cms-gem-daq-project/xhal/issues/111.

When the output file already exists, the function performs two sanity checks :
1. It validates the file header.
2. It ensure that the last character is EOL so that the new first line will not be merge with the previous last line.

If both checks pass, the new data are appended to the file. Therefore, it is possible to merge multiple OHs in a single result file.

### Types of changes

## How Has This Been Tested?

New tests have been written and pass :
``` bash
[lpetre@gem-daq reg_interface_gem]$ python -m unittest discover test
....
----------------------------------------------------------------------
Ran 4 tests in 0.019s

OK
```

Moreover the output file has been tested against the scripts from this PR : https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/201. The output plot corresponds to the generated pattern of the tests :

canv_GBTPhaseScanResults_aa

## Checklist:

### DIFF
--- a/python/reg_interface_gem/core/gbt_utils_extended.py
+++ b/python/reg_interface_gem/core/gbt_utils_extended.py
@@ -129,9 +129,24 @@ def saveGBTPhaseScanResults(filename, ohN, phasesBlob, nOfRepetitions=100):
             data.append([ohN, vfatN, phase, nOfRepetitions, nOfSuccesses])
 
     # Write the file
-    with open(filename, 'wb') as f:
-        f.write(header)
-        writer = csv.writer(f, delimiter = ' ')
+    # FIXME (Python 3) : The file cannot be opened in binary mode for the csv writer.
+    with open(filename, 'a+b') as f:
+        # Write header for empty (new) files
+        f.seek(0, 2) # Go to EOF
+        if f.tell() == 0:
+            f.write(header)
+
+        # Sanity checks
+        f.seek(0, 0) # Go to SOF
+        if f.readline() != header:
+            raise RuntimeError("The provided file ({0}) is either corrupted or not a GBT phase scan results file (invalid header).".format(filename))
+
+        f.seek(-1, 2) # Go to last character
+        if f.read(1) != '\n':
+            raise RuntimeError("The provided file ({0}) is either corrupted or not a GBT phase scan results file (no EOL before EOF).".format(filename))
+
+        # Write data
+        writer = csv.writer(f, delimiter = ' ', lineterminator = '\n')
         writer.writerows(data)
 
 def setPhase(cardName, ohN, vfatN, phase):
--- a/python/reg_interface_gem/test/data/test_saveGBTPhaseScanResults_AppendToFile.root
+++ b/python/reg_interface_gem/test/data/test_saveGBTPhaseScanResults_AppendToFile.root
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:1ca8c7bbc5600fa60d8a8ab421e89569c2a246b6909d54874f212148d1b6e06e
+size 22586
--- a/python/reg_interface_gem/test/data/test_saveGBTPhaseScanResults_NewFile.root
+++ b/python/reg_interface_gem/test/data/test_saveGBTPhaseScanResults_NewFile.root
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:b35efe89c3f5192b36cdcd813d4d90a45c4d7262dbcdf864df51e7ff6899fdda
+size 5684
--- a/python/reg_interface_gem/test/.gitattributes
+++ b/python/reg_interface_gem/test/.gitattributes
@@ -1,0 +1,1 @@
+data/** filter=lfs diff=lfs merge=lfs -text
--- a/python/reg_interface_gem/test/test_saveGBTPhaseScanResults.py
+++ b/python/reg_interface_gem/test/test_saveGBTPhaseScanResults.py
@@ -1,0 +1,74 @@
+import unittest
+
+from core.gbt_utils_extended import saveGBTPhaseScanResults
+
+import os
+from tempfile import mkstemp
+from ctypes import c_uint32
+from filecmp import cmp
+
+class TestSaveGBTPhaseScanResults(unittest.TestCase):
+
+    def setUp(self):
+        # Fill in a result array with a known pattern
+        self.phasesBlob = (c_uint32 * (24*16))()
+        for idx in range(len(self.phasesBlob)):
+            self.phasesBlob[idx] = idx
+
+        self.nOfRepetitions = len(self.phasesBlob) - 1
+
+        # Prepare a temp file
+        (self.fd, self.filename) = mkstemp()
+
+    def tearDown(self):
+        # Delete the temp file
+        os.close(self.fd)
+        os.remove(self.filename)
+
+    def testNewFile(self):
+        """
+        Test writing the GBT RX phase scan results of 1 OH to a file.
+        """
+
+        saveGBTPhaseScanResults(self.filename, 0, self.phasesBlob, self.nOfRepetitions)
+
+        # Compare with the reference file
+        referenceFile = os.path.dirname(os.path.abspath(__file__)) + "/data/test_saveGBTPhaseScanResults_NewFile.root"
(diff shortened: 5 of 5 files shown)


python/reg_interface_gem/test/data/test_saveGBTPhaseScanResults_AppendToFile.root
```diff
@@ -0,0 +1,1537 @@
+ohN/I:vfatN/I:phase/I:nRepetitions/I:nSuccesses/I
```
bdorney: What is the rational behind adding this file to the repo?

python/reg_interface_gem/test/data/test_saveGBTPhaseScanResults_NewFile.root
```diff
@@ -0,0 +1,385 @@
+ohN/I:vfatN/I:phase/I:nRepetitions/I:nSuccesses/I
```
bdorney: Same question as above.

bdorney (dismissed):

python/reg_interface_gem/test/data/test_saveGBTPhaseScanResults_AppendToFile.root
lpetre-ulb: These two files have been designed (even before writing the fix for the code) to be reference files for the tests, but also to be an example for developers.

I agree with you that these files are large, but I tried to minimize as much as possible external files in the tests. For example, corrupted files are generated on-the-fly. The tests could be considered more as integration tests than unit tests since files are written.

Also, these are the files which have been tested against your PR https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/201. (But the generated file is rigorously the same.)

mexanick (changes requested):
Please remove temporary `__init__.py`

lpetre-ulb:
I was trying to push the large objects for tests with git-lfs, but an error keeps occurring :
``` bash
[lpetre@gem-daq test]$ git push -f origin hotfix/save-gbt-phase-multiple-oh
Username for 'https://github.com': lpetre-ulb
Password for 'https://lpetre-ulb@github.com':
batch response: @lpetre-ulb can not upload new objects to public fork lpetre-ulb/xhal
error: failed to push some refs to 'https://github.com/lpetre-ulb/xhal.git'
```
It seems that either the parent repository needs to already have Git LFS object or I need write access to the parent repository as described on this [GitHub help pages](https://help.github.com/en/articles/collaboration-with-git-large-file-storage). What can I do ?

lpetre-ulb:
The two comments are addressed :
* `__init__.py` in `test/` is removed.
* Test object files are moved to Git LFS.

mexanick (dismissed):

mexanick:
please rebase on top of the current tip

lpetre-ulb:
> please rebase on top of the current tip

Rebased.

jsturdy (commented):
What is the expected/desired behaviour in the event that the file being written to *already* has phase data for the given set of GBTs?

bdorney:
> What is the expected/desired behaviour in the event that the file being written to _already_ has phase data for the given set of GBTs?

Good question.  But the use case I had in mind was that this was being called for multiple OH's in an `ohMask` by the same caller (e.g. how `testConnectivity.py` is working presently).  

If we wanted to handle this case I think we should read the file in; check if the OH being written exists, if it doens't just append to the file. If it does exist then I would replace all data for this OH, leaving any other data untouched, then write a new file.  In this way it only either adds data to the file or replaces the data for one OH.

python/reg_interface_gem/core/gbt_utils_extended.py
```diff
@@ -129,9 +129,24 @@ def saveGBTPhaseScanResults(filename, ohN, phasesBlob, nOfRepetitions=100):
...
+            f.write(header)
+
+        # Sanity checks
+        f.seek(0, 0) # Go to SOF
+        if f.readline() != header:
+            raise RuntimeError("The provided file ({0}) is either corrupted or not a GBT phase scan results file (invalid header).".format(filename))
+
+        f.seek(-1, 2) # Go to last character
+        if f.read(1) != '\n':
+            raise RuntimeError("The provided file ({0}) is either corrupted or not a GBT phase scan results file (no EOL before EOF).".format(filename))
```
bdorney, line 146: I would find some other method of checking if the file is how you want it to be.  For example a user might open this file and delete the last new line either manually or due to their `vim`/`emacs` settings (e.g. remove trailing whitespace) on opening/saving/closing the file.  So this could throw an unnecessary run time error.

If you want to do this I would find some other way to accommodate this check, a "simple" check could be:

```
#END GBT Phase Scan Results
```

And since it starts with a `#` the [`TTree::ReadFile`](https://root.cern.ch/doc/master/classTTree.html#a9c8da1fbc68221b31c21e55bddf72ce7) method will ignore this line.

Although it would be harder to parse this...some thought would be necessary.

jsturdy: However, in your use case outlined above, is there any user interaction with this file between different OHs? (trailing whitespace won't have an effect here, only lack of newline)

lpetre-ulb:
> > What is the expected/desired behaviour in the event that the file being written to _already_ has phase data for the given set of GBTs?
> 
> Good question. But the use case I had in mind was that this was being called for multiple OH's in an `ohMask` by the same caller (e.g. how `testConnectivity.py` is working presently).
> 
> If we wanted to handle this case I think we should read the file in; check if the OH being written exists, if it doens't just append to the file. If it does exist then I would replace all data for this OH, leaving any other data untouched, then write a new file. In this way it only either adds data to the file or replaces the data for one OH.

The current behavior is to simply append data. One use case could be to increase the total number of repetitions of a GBT phase scan if the program or the user decides there is not enough statistics after the first series of scan. This would, however, require patch in the parser.

I am not sure what is the best behavior, but the current one matches the current use case. If ever, this function would need to replace the data or throw an exception, the file would need to be entirely parsed at each function call.

python/reg_interface_gem/core/gbt_utils_extended.py
lpetre-ulb: Theses conditions have been designed as sanity checks, not bullet-proof checks. This precise condition ensures that the last character is a newline, as intended. Indeed the new data must be written to a new line a not merged with the last line of the previous data. If you or your text editor change the file in an incompatible way, the file is then considered corrupted (and it is corrupted even though it could be parsed).

bdorney: > However, in your use case outlined above, is there any user interaction with this file between different OHs? (trailing whitespace won't have an effect here, only lack of newline)

No within the scan and between OH's in the use case I outlined there is not.

bdorney:
> > > What is the expected/desired behaviour in the event that the file being written to _already_ has phase data for the given set of GBTs?
> > 
> > 
> > Good question. But the use case I had in mind was that this was being called for multiple OH's in an `ohMask` by the same caller (e.g. how `testConnectivity.py` is working presently).
> > If we wanted to handle this case I think we should read the file in; check if the OH being written exists, if it doens't just append to the file. If it does exist then I would replace all data for this OH, leaving any other data untouched, then write a new file. In this way it only either adds data to the file or replaces the data for one OH.
> 
> The current behavior is to simply append data. One use case could be to increase the total number of repetitions of a GBT phase scan if the program or the user decides there is not enough statistics after the first series of scan. This would, however, require patch in the parser.
> 
> I am not sure what is the best behavior, but the current one matches the current use case. If ever, this function would need to replace the data or throw an exception, the file would need to be entirely parsed at each function call.

I think for the current use case this is sufficient.  ~~If we want to increase flexibility and try to subvert problems then the save method could take a filename as an input with the current name as the default argument.  In this way it's up to the developer whose calling the function to ensure correct usage.~~ The function already allows an input file name argument.

~~Since this needs a rebase anyway...maybe @lpetre-ulb you could add that and it would then resolve this potential issue.~~ If you rebase this I'll approve it.  You've demonstrated this is suitable from my side.

lpetre-ulb:
Just rebased.

bdorney (approved):

bdorney:
And approved.

jsturdy (approved):

PR #114, opened 2019-04-30
Adding empty hostname verification

mexanick:
### Types of changes

## Motivation and Context

Required to prevent undefined behavior in case calling script is erroneus
## Checklist:

### DIFF
--- a/python/reg_interface_gem/scripts/gem_reg.py
+++ b/python/reg_interface_gem/scripts/gem_reg.py
@@ -18,10 +18,10 @@ if __name__ == '__main__':
         if (options.hostname):
             if (options.hostname == ""):
                 print 'Please provide non-empty hostname'
-                exit
+                sys.exit(os.EX_USAGE)
             if (rpc_connect(options.hostname)):
                 print '[Connection error] RPC connection failed'
-                exit
+                sys.exit(os.EX_USAGE)
         prompt=Prompt()
         prompt.execute(options.exe,args)
         sys.exit()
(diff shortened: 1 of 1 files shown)


python/reg_interface_gem/scripts/gem_reg.py
```diff
@@ -16,6 +16,9 @@
     if options.exe:
         parseXML()
         if (options.hostname):
+            if (options.hostname == ""):
+                print 'Please provide non-empty hostname'
+                exit
```
bdorney: would change line 21 to `exit(os.EX_USAGE)` with an import on `os` somewhere.

python/reg_interface_gem/scripts/gem_reg.py
```diff
@@ -16,9 +16,12 @@
     if options.exe:
         parseXML()
         if (options.hostname):
+            if (options.hostname == ""):
+                print 'Please provide non-empty hostname'
+                sys.exit(os.EX_USAGE)
             if (rpc_connect(options.hostname)):
```
jsturdy, line 22: Would it not make more sense to do the check inside this method?
If it's not something we can change, we should ask UW

mexanick: it's not something we want to change as the method comes from UW (I have a source code, but don't want to diverge). Also this doesn't go inline with quick fix. But in general - yes, I agree that it should simply throw the very same error as for bad hostname. I propose to merge this for the moment and request changes from UW.

bdorney (approved):

jsturdy (approved):
OK... my feelings on "quick fixes" are well known, but as this is the legacy branch anyway...

bdorney:
Is there a reason this is unmerged? @mexanick

PR #121, opened 2019-07-03
Bring core functionnality for the templated RPC calls

lpetre-ulb:
This PR provides the core functionality for the templated RPC calls.

## Description

Remotely callable RPC methods can now be registered to `rpcsvc` with the `xhal::rpc::registerMethod` function and called from the DAQ machine with the `xhal::rpc::call` function. The serialization and de-serialization is automatically done by the header library for the following types :

* `std::uint32_t`
* `std::vector<std::uint32_t>`
* `std::string`
* `std::vector<std::string>`

New types (`std::array`, `std::map` and custom types) will be added in a subsequent PR. Exceptions will also be take care of inside another PR.

Since this is the core of the templated RPC methods any further development rely on this PR. Therefore I would prefer this PR to be merged quickly with all but the outstanding comments addressed in a diffenrent PR. Of course the `rpc-templated-methods` must not be merged into `develop` until all of these comments are addressed.

### Types of changes

## Motivation and Context

This is the first PR of [this project](https://github.com/orgs/cms-gem-daq-project/projects/19).

## How Has This Been Tested?

It has been tested with the help of a dummy CPT7 module (reply with and/or log the method parameter(s)) and a ~10 lines C++ program on the DAQ machine.

## Checklist:

### DIFF
--- a/xhalcore/include/xhal/rpc/call.h
+++ b/xhalcore/include/xhal/rpc/call.h
@@ -1,0 +1,60 @@
+/*!
+ * \file
+ * \brief This file contains all the functions needed to call
+ * remotely a RPC method
+ *
+ * \author Laurent Pétré <lpetre@ulb.ac.be>
+ */
+
+#ifndef XHAL_RPC_CALL_H
+#define XHAL_RPC_CALL_H
+
+#include "xhal/rpc/common.h"
+#include "xhal/rpc/compat.h"
+#include "xhal/rpc/helper.h"
+
+#include "xhal/rpc/wiscrpcsvc.h" // move the header to xhal/extern/wiscrpcsvc.h ?
+
+#include <typeinfo>
+
+namespace xhal { namespace rpc {
+
+    /*!
+     * \brief Remotely call a RPC method
+     *
+     * The remote method called is defined by the template parameter
+     * \c Method. The arguments to give to the function are those from
+     * the \c Method::operator() signature and their types \b must be deducible.
+     */
+    template<typename Method,
+             typename... Args,
+             typename std::enable_if<std::is_base_of<xhal::rpc::Method, Method>::value, int>::type = 0
+            >
+    helper::functor_return_t<Method> call(wisc::RPCSvc &connection, Args&&... args)
+    {
+        // The wisc::RPCMsg method name is taken from the typeid
+        // This is implementation dependent but g++ and clang++
+        // follow the same convention
+        wisc::RPCMsg request(std::string(abiVersion) + "." + typeid(Method).name());
+        MessageSerializer query{&request};
+
+        // Type conversion from args to serializable types
+        // must be performed in the same statement in order to
+        // make use of lifetime extension
+        query << helper::get_forward_as_tuple<Method>()(args...);
+
+        // Remote call
+        const wisc::RPCMsg response = connection.call_method(request);
+
+        // The RPC method can return a void so the void_holder is required
+        compat::void_holder<helper::functor_return_t<Method>> return_v;
+
+        MessageDeserializer reply{&response};
+        reply >> return_v;
+
+        return return_v.get();
+    }
+
+}}
+
+#endif
--- a/xhalcore/include/xhal/rpc/common.h
+++ b/xhalcore/include/xhal/rpc/common.h
@@ -1,0 +1,313 @@
+/*!
+ * \file
+ * \brief Contains the classes required for defining remotely callable RPC methods
+ *
+ * \author Laurent Pétré <lpetre@ulb.ac.be>
+ */
+
+#ifndef XHAL_RPC_COMMON_H
+#define XHAL_RPC_COMMON_H
+
+#include "xhal/rpc/compat.h"
+
+#include <cstdint>
+#include <string>
+#include <vector>
+
+#include "xhal/rpc/wiscRPCMsg.h" // move the header to "xhal/extern/wiscRPCMsg.h" ?
+
+namespace xhal { namespace rpc {
+
+    /*!
+     * \brief Defines the templated RPC ABI version
+     */
+    static constexpr const char* abiVersion = "v1";
+
+    /*!
+     * \brief Class whose all remotely callable RPC method must inherit from
+     *
+     * The required inheritance is used as a compile time check so a developer
+     * cannot remotely call a local function by mistake.
+     */
+    struct Method
+    {
+        /*!
+         * \brief The operator call must be define <b>once and only once</b> per
+         * RPC method.
+         *
+         * This templated operator declaration is only shown as an example and
+         * emphasizes the need of defining it in child classes.
+         *
+         * \warnng The call operator \b must be defined as \c const.
+         */
+        template<typename R, typename... Args> R operator()(Args...) const;
+    };
+
+    /*!
+     * \brief Base of the \c MessageSerializer and \c MessageDeserializer classes
+     *
+     * \c MessageBase provides the key index tracking functionnality which
+     * is mandatory for serialization.
+     */
+    class MessageBase {
+
+        /*!
+         * \brief Index to the next free/unread key
+         */
+        uint32_t _keyIdx = 0;
+
+      protected:
+
+        /*
+         * \brief Returns the next free/unread key
+         */
+        inline uint32_t dispenseKey() { return _keyIdx++; }
+
+    };
+
+    /*!
+     * \brief This class serializes parameters into a \c wisc::RPCMsg
+     */
+    class MessageSerializer : public MessageBase
+    {
+
+    protected:
+
+        wisc::RPCMsg *m_wiscMsg;
+
+        /*
+         * \brief Supresses implicit type conversions
+         *
+         * Every type not defined hereunder is delete by this templated function.
+         * It aims at enforcing maximum type compatibility with the UW RPC API by
+         * remembering the developer that she/he can transmit defined types over the
+         * network.
+         */
+        template<typename T> void save(T) = delete;
+
+        /*!
+         * \brief Adds a \c std::uint32_t to the message
+         */
+        inline void save(const std::uint32_t value) {
+            m_wiscMsg->set_word(std::to_string(dispenseKey()), value);
+        }
+
+        /*!
+         * \brief Adds a \c std::vector<std::uint32_t> to the message
+         */
+        inline void save(const std::vector<std::uint32_t> &value) {
+            m_wiscMsg->set_word_array(std::to_string(dispenseKey()), value);
+        }
+
+        /*!
+         * \brief Adds a \c std::string to the message
+         */
+        inline void save(const std::string &value) {
+            m_wiscMsg->set_string(std::to_string(dispenseKey()), value);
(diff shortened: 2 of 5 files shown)


xhalcore/include/xhal/rpc/helper.h
```diff
@@ -0,0 +1,114 @@
...
+ */
+
+#ifndef XHAL_RPC_HELPER_H
+#define XHAL_RPC_HELPER_H
+
+#include <tuple>
+
+namespace xhal { namespace rpc { namespace helper {
+
+    /*
```
lmoureaux: `/*!`

xhalcore/include/xhal/rpc/helper.h
```diff
@@ -0,0 +1,114 @@
...
+    /*!
+     * \brief Helper function to forward as a tuple matching the functor signature
+     */
+    template<typename Obj>
+    constexpr inline auto get_forward_as_tuple() {
+        return functor_traits<decltype(&Obj::operator())>::forward_as_tuple();
+    }
+
+    /*!
+     * \brief Checks wether the template parameter \c T is a \c std::tuple
```
lmoureaux: wHether

lmoureaux (commented):
Looks good.
Does the code pass `-Werror -Wall -Wextra -pedantic -pedantic-errors` on modern `gcc` and `clang`?

lpetre-ulb:
> Does the code pass `-Werror -Wall -Wextra -pedantic -pedantic-errors` on modern `gcc` and `clang`?

`gcc 7.3.1` and `clang 5.0.1` produce some warnings in header files outside the scope of this PR (and even outside the GEM code). Therefore I couldn't enable `-Werror` and `-pedantic-errors`.

xhalcore/include/xhal/rpc/helper.h
lpetre-ulb: Woops...

lmoureaux (approved):

jsturdy:
I'm just thinking about the structure, where currently we have a package built for "`arm`"  and a "`core`" package (in quotes because I imagine that the expectation would be that the "`arm`" part would also be what would be needed inside the `docker` container emulator).
It probably doesn't matter now, but I had to think about it because the `registerMethod` function belongs only in the environment where the `ctp7_modules` are built (currently the "`arm`" part).
We are currently packaging the `x86_64` package and `devel` package, and had  bundled the `arm` lib with the `devel` package to provide all the parts necessary for building modules for the CTP7, so we could envision the same for this (as I *think* we'll still need an `xhal` library on the CTP7/`docker` container, which means the `devel` package should still provide the `arm` compatible library for building)

Anyway, those are my rambling thoughts

xhalcore/include/xhal/rpc/compat.h
```diff
@@ -0,0 +1,146 @@
+/*!
+ * \file
+ * \brief Compatibility functionnalities for C++11
```
mexanick: functioNalities ;)

xhalcore/include/xhal/rpc/compat.h
```diff
@@ -0,0 +1,146 @@
...
+ *
+ * This file includes all the functionnalities that could advantagously be
```
mexanick: functioNalities & advantagEously

xhalcore/include/xhal/rpc/compat.h
```diff
@@ -0,0 +1,146 @@
...
+ * replaced with newer versions of C++, e.g. C++14 and C++17.
+ *
+ * Since the content of this file implements missing functionnalties in the
```
mexanick: functioNalities

xhalcore/include/xhal/rpc/compat.h
```diff
@@ -0,0 +1,146 @@
...
+     * \brief Helper making an index sequence from \c 0 to \c N-1
+     *
+     * Remind that an index sequence is the non-type template parameter of a
+     * specialization of the \c index_sequence template class.
+     */
+    template<std::size_t N>
+        using make_index_sequence = typename index_sequence_gen<N>::type;
+
+    /*!
+     * \brief Calls a function with argumments from a \c std::tuple
```
mexanick: arguMents

mexanick (commented):
Just few typos to fix. Considering @jsturdy's comment: the `core` functionality supposed to be used both on host and CTP7, while the `arm` - only on the card, but this division is conditional and based on use-case logic. In case of embedding the `arm` code into the container, the logic will stay the same. One may want to change the name to be a more appropriate one, but I honestly don't see much contradiction in using the current.

xhalcore/include/xhal/rpc/call.h
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+#ifndef XHAL_RPC_CALL_H
+#define XHAL_RPC_CALL_H
+
+#include "xhal/rpc/common.h"
+#include "xhal/rpc/compat.h"
+#include "xhal/rpc/helper.h"
+
+#include "xhal/rpc/wiscrpcsvc.h" // move the header to xhal/extern/wiscrpcsvc.h ?
```
jsturdy, line 16: Better, remove these files from `xhal` and use a `devel` package of `rpcsvc` (also for `wiscRPCMsg.h`)

lpetre-ulb: I fully agree with you, especially since these files are copy-paste in different locations. Though I consider it out of the scope of this first templated RPC PR. Also who is responsible for the creation of the `wiscrpcsvc` package? I would have expected UW to provide its `devel` version...

xhalcore/include/xhal/rpc/compat.h
lpetre-ulb: Whoops... I must definitively enable spelling correction in my code text editor. :)

lpetre-ulb:
> I'm just thinking about the structure, where currently we have a package built for "`arm`" and a "`core`" package (in quotes because I imagine that the expectation would be that the "`arm`" part would also be what would be needed inside the `docker` container emulator).

As you suggested to me two weeks ago after the DAQ meeting, I will try to use the "`core`" package inside the `docker` image since it provides all the functionalities embedded within the "`arm`" package and is already built for x86_64.  think the only difference is that `libxhal.so` in the "`arm`" package provides only a subset of the symbols present in "`core`".

> It probably doesn't matter now, but I had to think about it because the `registerMethod` function belongs only in the environment where the `ctp7_modules` are built (currently the "`arm`" part).
> We are currently packaging the `x86_64` package and `devel` package, and had bundled the `arm` lib with the `devel` package to provide all the parts necessary for building modules for the CTP7, so we could envision the same for this (as I _think_ we'll still need an `xhal` library on the CTP7/`docker` container, which means the `devel` package should still provide the `arm` compatible library for building)

If I'm not mistaken all the headers files are already included in the `devel` package. Therefore the templated RPC functionalities, which are header only, will be installed along with the development package. Of course `registerMethod` should not be called from the DAQ machine but the developer will be warned soon enough if she/he tries to use it that way. 

For the packaging the library header files are usually installed once for all the architectures since they should be the same. Currently this path is followed. 

I think the current packaging works, but final packaging will be done as the last step.

> Just few typos to fix.

Fixed.

> Considering @jsturdy's comment: the `core` functionality supposed to be used both on host and CTP7, while the `arm` - only on the card, but this division is conditional and based on use-case logic. In case of embedding the `arm` code into the container, the logic will stay the same. One may want to change the name to be a more appropriate one, but I honestly don't see much contradiction in using the current.

Since the `xHAL` library is really at the interface between the CTP7 and the DAQ machine I think it is expected to get different functionalities depending on the target system. Sure "`arm`" might not be the best name if one consider the GLIB, but that does not shock me. What I don't like though is that the `libxhal.so` library provides different symbols and functionalities for different targets. I would really prefer to split it in N (2 currently) libraries. For example `libxhal_core.so` and `libxhal_interface.so` where the second one is only provided for the DAQ machine.

xhalcore/include/xhal/rpc/call.h
bdorney: @lpetre-ulb the `wiscrpcsvc` package is shipped with the linux image of all CTP7's, see:

```
/mnt/image-persist/image_doc/rpcsvc/
```

xhalcore/include/xhal/rpc/common.h
```diff
@@ -0,0 +1,313 @@
...
+#include <vector>
+
+#include "xhal/rpc/wiscRPCMsg.h" // move the header to "xhal/extern/wiscRPCMsg.h" ?
+
+namespace xhal { namespace rpc {
+
+    /*!
+     * \brief Defines the templated RPC ABI version
+     */
+    static constexpr const char* abiVersion = "v1";
```
bdorney, line 24: Is there someway this could be picked up at build time?  E.g. similar to how `tag2rel.sh` works?

lpetre-ulb: I guess it should be possible, but what value would you want to use?

This is the templated RPC ABI version, i.e. how data are serialized and exceptions handled, so there is very little chance for it to ever change; maybe when the ATCA `rpcsvc`-like comes, but there is nothing more unsure.

xhalcore/include/xhal/rpc/call.h
lpetre-ulb: @bdorney, I cannot find the `wiscrpsvc` package in `/mnt/image-persist/image_doc/rpcsvc/` : 

``` bash
eagle63:~$ ll /mnt/image-persist/image_doc/rpcsvc/
-rw-r--r--    1 root     root          6151 Jun 22  2018 client_dev.tbz2
-rw-r--r--    1 root     root          7489 Jun 22  2018 module_dev.tbz2
```

nor inside the TAR archives. But if it is present inside the Linux image I would expect UW to create the corresponding `devel` version. If not the cleanest way would be to patch the [`rpcsvc`](https://github.com/cms-gem-daq-project/rpcsvc) repo (yes, once again...)

xhalcore/include/xhal/rpc/common.h
bdorney: Let me ask a different question; what is this constant `abiVersion` used for? Like in the current architecture we need to specify a version when registering a method; is this the use case...or...?

Probably my comment comes from my ignorance on what this guy is used for.  It would be better to understand that before asking for changes ;)

lmoureaux: It is the version of the communication protocol between the PC and the CTP7, that is the layout of the underlying `wiscrpcsvc` messages (key names and types). If an incompatible update of the protocol is ever needed, it will be used to prevent communication between binaries using different versions (technically the CTP7 modules could support several protocol versions).

At the moment the code doesn't mess up with the way modules are loaded. One still needs to write the module version in the source.

lmoureaux:
Are the comments blocking or can this PR be merged?

bdorney:
> Are the comments blocking or can this PR be merged?

My comment certainly doesn't block; probably ping @mexanick and @jsturdy

jsturdy (approved):

PR #123, opened 2019-07-11
Add more types to the templated RPC

lpetre-ulb:
## Description

This PR adds support for new types in the templated RPC methods:

* `std::array<T, N>` where `T` is a serializable type
* `std::map<std::uint32_t, T>` where `T` is a serializable type
* `std::map<std::string, T>` where `T` is a serializable type
* Custom types

### Types of changes

## Motivation and Context

This is the continuation of the the [previous PR](https://github.com/cms-gem-daq-project/xhal/pull/121) about the templated RPC methods.

## How Has This Been Tested?

It has been tested with the help of a dummy CPT7 module (reply with and/or log the method parameter(s)) and a ~10 lines C++ program on the DAQ machine.

## Checklist:

### DIFF
--- a/xhalcore/include/xhal/rpc/common.h
+++ b/xhalcore/include/xhal/rpc/common.h
@@ -79,15 +79,22 @@ namespace xhal { namespace rpc {
 
         wisc::RPCMsg *m_wiscMsg;
 
-        /*
-         * \brief Supresses implicit type conversions
+        /*!
+         * \brief Serializes custom types if possible or else supresses implicit type conversions
          *
-         * Every type not defined hereunder is delete by this templated function.
-         * It aims at enforcing maximum type compatibility with the UW RPC API by
-         * remembering the developer that she/he can transmit defined types over the
-         * network.
+         * Every type not defined hereunder is taken care of by this templated function.
+         * The function serves two purposes:
+         *
+         * 1. It delegates the serialization to a well-known function.
+         * 2. It aims at enforcing maximum type compatibility with the UW RPC API by
+         *    remembering the developer that she/he can transmit defined types over the
+         *    network.
          */
-        template<typename T> void save(T) = delete;
+        template<typename T> inline void save(const T &t) {
+            // This const_cast is safe when the API is used as intented
+            // More precisely when the object t is modified only with the operator&
+            serialize(*this, const_cast<T &>(t));
+        }
 
         /*!
          * \brief Adds a \c std::uint32_t to the message
@@ -217,7 +224,7 @@ namespace xhal { namespace rpc {
          * \brief Allows to serialize data into the message with a natural interface
          */
         template<typename T>
-        MessageSerializer & operator<<(const T &t) {
+        inline MessageSerializer & operator<<(const T &t) {
             this->save(t);
             return *this;
         }
@@ -230,7 +237,7 @@ namespace xhal { namespace rpc {
          * function.
          */
         template<typename T>
-        MessageSerializer & operator&(const T &t) {
+        inline MessageSerializer & operator&(const T &t) {
             this->save(t);
             return *this;
         }
@@ -250,15 +257,20 @@ namespace xhal { namespace rpc {
 
         const wisc::RPCMsg *m_wiscMsg;
 
-        /*
-         * \brief Supresses implicit type conversion
+        /*!
+         * \brief Deserializes custom types if possible or else supresses implicit type conversion
          *
-         * Every type not defined hereunder is delete by this templated function.
-         * It aims at enforcing maximum type compatibility with the UW RPC API by
-         * remembering the developer that she/he can transmit defined types over the
-         * network.
+         * Every type not defined hereunder is taken care of by this templated function.
+         * The function serves two purposes:
+         *
+         * 1. It delegates the deserialization to a well-known function.
+         * 2. It aims at enforcing maximum type compatibility with the UW RPC API by
+         *    remembering the developer that she/he can transmit defined types over the
+         *    network.
          */
-        template<typename T> void load(T) = delete;
+        template<typename T> inline void load(T &t) {
+            serialize(*this, t);
+        }
 
         /*!
          * \brief Retrieves a \c std::uint32_t from the message
@@ -363,7 +375,7 @@ namespace xhal { namespace rpc {
                  typename... Args,
                  typename std::enable_if<I == sizeof...(Args), int>::type = 0
                 >
-        inline void load(const std::tuple<Args...> &) { }
+        inline void load(std::tuple<Args...> &) { }
 
     public:
 
@@ -378,7 +390,7 @@ namespace xhal { namespace rpc {
          * \brief Allows to deserialiaze data from the message with a natural interface
          */
         template <typename T>
-        MessageDeserializer & operator>>(T &t) {
+        inline MessageDeserializer & operator>>(T &t) {
             this->load(t);
             return *this;
         }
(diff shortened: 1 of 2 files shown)


xhalcore/include/xhal/rpc/common.h
```diff
@@ -126,6 +128,24 @@ namespace xhal { namespace rpc {
...
 
+        /*!
+         * \brief Adds a \c std::map<std::uint32_t, T> to the message where \c T is a serializable type
+         */
+        template<typename T> inline void save(const std::map<std::uint32_t, T> &value) {
+            // The first RPC key stores the std::map keys
+            // This is required to know the std::map size at deserialization
+            const auto keysKey = dispenseKey();
+
+            std::vector<std::uint32_t> keys{};
```
lmoureaux, line 146: Use `keys.reserve(value.size())` to reduce the number of memory allocations.

xhalcore/include/xhal/rpc/common.h
```diff
@@ -259,6 +279,18 @@ namespace xhal { namespace rpc {
...
         }
 
+        /*!
+         * \brief Retrieves a \c std::map<std::uint32_t, T> from the message where \c T is a serializable type
+         */
+        template<typename T> inline void load(std::map<std::uint32_t, T> &value) {
+            const auto keys = m_wiscMsg->get_word_array(std::to_string(dispenseKey()));
+
+            for (const auto & key: keys) {
+                T val; this->load(val);
```
lmoureaux: Style: line break after `;`

xhalcore/include/xhal/rpc/common.h
```diff
@@ -146,6 +146,24 @@ namespace xhal { namespace rpc {
...
 
+        /*!
+         * \brief Adds a \c std::map<std::string, T> to the message where \c T is a serializable type
+         */
+        template<typename T> inline void save(const std::map<std::string, T> &value) {
+            // The first RPC key stores the std::map keys
+            // This is required to know the std::map size at deserialization
+            const auto keysKey = dispenseKey();
+
+            std::vector<std::string> keys{};
```
lmoureaux, line 165: `reserve`

xhalcore/include/xhal/rpc/common.h
```diff
@@ -291,6 +309,18 @@ namespace xhal { namespace rpc {
...
         }
 
+        /*!
+         * \brief Retrieves a \c std::map<std::string, T> from the message where \c T is a serializable type
+         */
+        template<typename T> inline void load(std::map<std::string, T> &value) {
+            const auto keys = m_wiscMsg->get_string_array(std::to_string(dispenseKey()));
+
+            for (const auto & key: keys) {
+                T val; this->load(val);
```
lmoureaux: Line break

xhalcore/include/xhal/rpc/common.h
```diff
@@ -91,7 +91,9 @@ namespace xhal { namespace rpc {
          *    network.
          */
         template<typename T> void save(const T &t) {
-            t.save(*this);
+            // This const_cast is safe when the API is properly used
+            // When data are saved the operator& only reads and never writes
+            const_cast<T &>(t).serialize(*this);
```
lmoureaux: I'm not sure how this API could be used to serialize a type defined by a library (eg `std::valarray`). I get the impression that types not supported within `Message[De]serializer` are second-class citizens.

In any case some documentation about custom types would be welcome ;)

lmoureaux (changes requested):
The first two commits are almost perfect (small perf and style comments), but I don't really understand how custom types are supposed to work. Requesting more documentation ;)

xhalcore/include/xhal/rpc/common.h
lpetre-ulb: Yes, good point. :-)

xhalcore/include/xhal/rpc/common.h
lpetre-ulb: I like this style when it is not possible to perform initialization at declaration. But ok, I'll change it.

xhalcore/include/xhal/rpc/common.h
lpetre-ulb: Indeed types defined in a library are not currently supported. Since I really don't like the `const_casst` (if even it is safe when the API is properly used) I thought about another implementation. I'll fix that and include an example (for `std::array<T, N>`) as an example in the  documentation.

lpetre-ulb:
Support for types defined in libraries added as well as one custom type serialization example.

xhalcore/include/xhal/rpc/common.h
```diff
@@ -391,13 +403,66 @@ namespace xhal { namespace rpc {
...
+     *
+     *     // The intrusive version is implemented as a new member function
+     *     // which takes a message as parameter (i.e. the serializer or deserializer)
+     *     template<class Message> inline void serialize(Message & msg) {
+     *         msg & x & y;
+     *     }
+     * };
+     *
+     * // The non-intrusive version allows to serialize objects defined in a library
+     * // Simply define the serialize function in the xhal::rpc namespace with two parameters
```
lmoureaux: > in the `xhal::rpc` namespace

Is it possible to leverage [ADL](https://en.cppreference.com/w/cpp/language/adl)? I think it's better than defining the function in another `namespace`, but the rules can be... confusing. (One can't use ADL for `std` classes, but it would be possible for custom ones.)

xhalcore/include/xhal/rpc/common.h
```diff
@@ -391,13 +403,66 @@ namespace xhal { namespace rpc {
...
+     * } }
+     * \endcode
+     *
+     * \warning In order to work as intended the \c serialize functions \b MUST modify
+     * the object only with the \c operator&
+     *
+     */
+    template<typename Message, typename T, std::size_t N>
+    inline void serialize(Message &msg, std::array<T, N> &value) {
+        for (auto & elem: value) {
```
lmoureaux, line 463: Maybe add the following comment:
```cpp
// Since the size of the array is known at compile time (and part of the signature),
// we don't need to serialize it.
```

lmoureaux (changes requested):
Nice improvement with the use of an external function :+1: Got two small comments.

xhalcore/include/xhal/rpc/common.h
lpetre-ulb: Yes, it is. But ADL with templates is very tricky...

If you define the `serialize` function in the same namespace as the type you want to serialize the function is found and used. (Let's wait C++20 to allow support for the `std` classes...) You can also define a file scope `serialize` function, but it must come *before* the `Message[De]serializer` classes: only ADL lookup is performed in the instantiation context. So if the types is defined in the global namespace it works.

So yes possible, but subtle. Do you want me to relax the constraints in the documentation? For e.g.
> in the xhal::rpc namespace or the namespace where the type is defined

xhalcore/include/xhal/rpc/common.h
```diff
@@ -75,15 +79,22 @@ namespace xhal { namespace rpc {
 
         wisc::RPCMsg *m_wiscMsg;
```
jsturdy, line 80: Any reason this isn't a member of the base class?
Also, my usual question, does it make sense to have this a smart pointer, or are there no risks of leaks?

lpetre-ulb: > Any reason this isn't a member of the base class?

Yes, this member is `const` in the deserializer (see line 258).

> Also, my usual question, does it make sense to have this a smart pointer, or are there no risks of leaks?

Not really; this is a raw pointer given by the `rpcsvc` daemon which manages its lifetime. And when used in `xhal::rpc::call` the address of the local object is used.

jsturdy: > > Any reason this isn't a member of the base class?
> 
> Yes, this member is `const` in the deserializer (see line 258).

Ah, yes, I elided the `const` part there.

> 
> > Also, my usual question, does it make sense to have this a smart pointer, or are there no risks of leaks?
> 
> Not really; this is a raw pointer given by the `rpcsvc` daemon which manages its lifetime. And when used in `xhal::rpc::call` the address of the local object is used.

OK

jsturdy (approved):

xhalcore/include/xhal/rpc/common.h
lmoureaux: The proposed wording is more clear. I think we should use ADL for our own types and `xhal::rpc` for everything else.

jsturdy: @lmoureaux, is this the only thing remaining from "changes requested"?

xhalcore/include/xhal/rpc/common.h
lmoureaux: @jsturdy also this one

lpetre-ulb:
@lmoureaux's comments addressed.

jsturdy (approved):
One minor comment that can be addressed in a future PR is the style for namespaces in header files.
Here you have:
```c++
namespace xhal { namespace rpc {
/// content
}}
```
and in @lmoureaux's PR (#125) he uses:
```c++
namespace xhal::rpc::helper
/// content
}}}
```

Neither of these conform with the current style of :
```c++
namespace xhal {
  namespace rpc {
/// content
  }
}
```

I don't have a strong (enough to lay down the law) preference for either (though I will continue to use the latter in `cmsgemos`), but I do have a requirement for consistency, so if someone has a strong preference one way or the other (@mexanick, @lmoureaux, @lpetre-ulb, @bdorney), please raise it, otherwise, I will adopt the same convention as is used in our other code.

lpetre-ulb:
> One minor comment that can be addressed in a future PR is the style for namespaces in header files.

I agree that we must use a uniform style; I tried to follow the [cmsgemos CONTRIBUTING guideline](https://github.com/cms-gem-daq-project/cmsgemos/blob/develop/.github/CONTRIBUTING.md), but it is not written there. Since one or two small PRs will be required to bring the last touch I also think it can be delayed until then. 

> I don't have a strong (enough to lay down the law) preference for either (though I will continue to use the latter in `cmsgemos`), but I do have a requirement for consistency, so if someone has a strong preference one way or the other (@mexanick, @lmoureaux, @lpetre-ulb, @bdorney), please raise it, otherwise, I will adopt the same convention as is used in our other code.

My only concern with the current style is that deeply nested namespaces increase the indentation by a lot, especially if one indentation is 4 spaces.

jsturdy:
> > I don't have a strong (enough to lay down the law) preference for either (though I will continue to use the latter in `cmsgemos`), but I do have a requirement for consistency, so if someone has a strong preference one way or the other (@mexanick, @lmoureaux, @lpetre-ulb, @bdorney), please raise it, otherwise, I will adopt the same convention as is used in our other code.
> 
> My only concern with the current style is that deeply nested namespaces increase the indentation by a lot, especially if one indentation is 4 spaces.

There are those (I am not one) who mandate a style of *not* indenting namespaces..., for me indentation adds visual scope information, and as we live and code in the 21st century, I am *not* concerned with strict 80 character line limits (though the way `github` displays files with ample unused whitespace on either side of the main text, *could* be a reason to adopt some guideline...)

mexanick:
I'd support the most populated namespace identation style aka 
```cpp
namespace xhal {
  namespace rpc {
/// content
  }
}
```

mexanick (approved):

PR #124, opened 2019-07-11
Implementing callers for SCAADC readouts

mexanick:
## Description

Caller wrapper for SCA ADC readout
Requires clarification on return documentation - see https://github.com/cms-gem-daq-project/ctp7_modules/issues/142
### Types of changes

## How Has This Been Tested?

Not yet tested, WIP
## Checklist:

### DIFF
--- a/xhalcore/src/common/rpc_manager/sca.cc
+++ b/xhalcore/src/common/rpc_manager/sca.cc
@@ -1,0 +1,162 @@
+#include "xhal/rpc/sca.h"
+
+DLLEXPORT uint32_t readSCAADCSensor(const uint32_t ohMask, const uint32_t ch, uint32_t* result)
+{
+    req = wisc::RPCMsg("amc.readSCAADCSensor");
+
+    req.set_word("ohMask", ohMask);
+    req.set_word("ch", ch);
+
+    wisc::RPCSvc* rpc_loc = getRPCptr();
+
+    try {
+        rsp = rpc_loc->call_method(req);
+    }
+    STANDARD_CATCH;
+
+    if (rsp.get_key_exists("error")) {
+        printf("Caught an error: %s\n", (rsp.get_string("error")).c_str());
+        return 1;
+    }
+    const uint32_t size = count_1bits(ohMask);
+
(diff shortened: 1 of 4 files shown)


bdorney:
Is this `WIP`? Review was requested but I see `In Progress` tag and `WIP` in title; please advise.

mexanick:
Yes, review with suggestion on documenting the resulting array (from @ram1123) is expected. Of course, other comments are also welcome.

mexanick:
@ram1123 please provide feedback

xhalcore/src/common/rpc_manager/sca.cc
```diff
@@ -0,0 +1,162 @@
...
+    STANDARD_CATCH;
+
+    if (rsp.get_key_exists("error")) {
+        printf("Caught an error: %s\n", (rsp.get_string("error")).c_str());
+        return 1;
+    }
+    const uint32_t size = count_1bits(ohMask);
+
+    if (rsp.get_key_exists("data")){
+        ASSERT(rsp.get_word_array_size("data") == size);
```
ram1123, line 24: @mexanick : Are you sure that number of 1's in `ohMask` is always equal to 1's in `data`. Lets say if there is only one OH is unmasked its its **size** will be 1 but data might have more than 1 ones.

I don't have much idea so I might be wrong.

ram1123 (commented):
After this small comments. From my side its fine.

xhalcore/src/common/rpc_manager/sca.cc
jsturdy: Reading a single sensor will only ever return one value per unmasked OH, so in this case it's "fine".
For L48, it's multiplied by 5, a magic number which presumably indicates the number of temperature sensors...
Likewise in L80, L112, and L144, the magic numbers 6, 3, and 14
I don't like this, but if it works, then it serves its purpose for now, and this code *won't* be migrated into the templated framework

jsturdy (approved):
If it works, I'll sign off, as this is *just* for the legacy "get-something-working-now" branch

ram1123 (approved):

PR #125, opened 2019-07-20
Exception handling in template RPC calls

lmoureaux:
## Description

Handle exceptions thrown by RPC methods and the Wisconsin RPC library. Turn them into two exceptions types at the calling site. For exceptions thrown on the remote side, a stack trace can be recorded and sent back (optional).

This PR adds a `.cpp` file that should be compiled into its a library. It is required on all architectures. The build system has *not* been modified to include it.

### Example exception info

I triggered an exception by doing `.at(-1)` on a `std::vector`. The message is:
```
remote error: std::out_of_range: vector::_M_range_check: __n (which is 18446744073709551615) >= this->size() (which is 10)
```

*[feature was dropped]*
The backtrace as retrieved from the thrown exception is (relevant frames are 5-7):
```
Stack trace (most recent call last):
#17   Object "[0xffffffffffffffff]", at 0xffffffffffffffff, in 
#16   Object "wiscrpcsvc-server.pc/src/wiscrpcsvc-server.pc-build/rpcsvc", at 0x55b4f5f22249, in _start
#15   Object "/lib/x86_64-linux-gnu/libc.so.6", at 0x7f01aa1fdb96, in __libc_start_main
#14   Object "wiscrpcsvc-server.pc/src/wiscrpcsvc-server.pc-build/rpcsvc", at 0x55b4f5f228c9, in main
#13   Object "wiscrpcsvc-server.pc/src/wiscrpcsvc-server.pc-build/rpcsvc", at 0x55b4f5f22b25, in handle_client_connection(int, sockaddr_in&, unsigned int, int&)
#12   Object "wiscrpcsvc-server.pc/src/wiscrpcsvc-server.pc-build/rpcsvc", at 0x55b4f5f23779, in run_client(int)
#11   Object "wiscrpcsvc-server.pc/src/wiscrpcsvc-server.pc-build/rpcsvc", at 0x55b4f5f25fb5, in ModuleManager::invoke_method(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, wisc::RPCMsg*, wisc::RPCMsg*)
#10   Object "/home/louis/Work/GEM/superbuild/gemsuperbuild/install/pc/usr/lib/rpcmodules/memory.so", at 0x7f01a93a5212, in void xhal::rpc::invoke<Memory::Read, 0>(wisc::RPCMsg const*, wisc::RPCMsg*)
#9    Object "/home/louis/Work/GEM/superbuild/gemsuperbuild/install/pc/usr/lib/rpcmodules/memory.so", at 0x7f01a93a5aad, in xhal::rpc::compat::void_holder<std::vector<unsigned int, std::allocator<unsigned int> > > xhal::rpc::compat::tuple_apply<std::vector<unsigned int, std::allocator<unsigned int> >, Memory::Read, unsigned int, unsigned int, 0>(Memory::Read&&, std::tuple<unsigned int, unsigned int> const&)
#8    Object "/home/louis/Work/GEM/superbuild/gemsuperbuild/install/pc/usr/lib/rpcmodules/memory.so", at 0x7f01a93a674f, in decltype (((forward<Memory::Read>)({parm#1}))((get<0ul>)({parm#2}), (get<1ul>)({parm#2}))) xhal::rpc::compat::tuple_apply_impl<Memory::Read, unsigned int, unsigned int, 0ul, 1ul>(Memory::Read&&, std::tuple<unsigned int, unsigned int> const&, xhal::rpc::compat::index_sequence<0ul, 1ul>)
#7    Object "/home/louis/Work/GEM/superbuild/gemsuperbuild/install/pc/usr/lib/rpcmodules/memory.so", at 0x7f01a93a34ed, in Memory::Read::operator()(unsigned int, unsigned int) const
#6    Object "/home/louis/Work/GEM/superbuild/gemsuperbuild/install/pc/usr/lib/rpcmodules/memory.so", at 0x7f01a93a4a22, in std::vector<unsigned int, std::allocator<unsigned int> >::at(unsigned long)
#5    Object "/home/louis/Work/GEM/superbuild/gemsuperbuild/install/pc/usr/lib/rpcmodules/memory.so", at 0x7f01a93a50d3, in std::vector<unsigned int, std::allocator<unsigned int> >::_M_range_check(unsigned long) const
#4    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6", at 0x7f01aa873854, in 
#3    Object "../install/pc/usr/lib/libxhal_rpc.so", at 0x7f01ab4b7a3e, in __cxa_throw
#2    Object "../install/pc/usr/lib/libxhal_rpc.so", at 0x7f01ab4b789d, in xhal::rpc::helper::recordStackTrace()
#1    Object "../install/pc/usr/lib/libxhal_rpc.so", at 0x7f01ab4b834d, in backward::StackTraceImpl<backward::system_tag::linux_tag>::load_here(unsigned long)
#0    Object "../install/pc/usr/lib/libxhal_rpc.so", at 0x7f01ab4b9d15, in unsigned long backward::details::unwind<backward::StackTraceImpl<backward::system_tag::linux_tag>::callback>(backward::StackTraceImpl<backward::system_tag::linux_tag>::callback, unsigned long)
```
Line numbers and even code snippets could be obtained by linking against `libdwarf`

### How it works

Exceptions are caught on the server side (CTP7, in `invoke`) and an error message is generated there. The exact type of the exception is determined using a low-level C++ API (defined by the Itanium ABI). This information is sent back to the caller using the `error` and `type` keys. The `call` function checks for the keys and throws an exception with the relevant information.

In addition, code in the `call` function is now protected against exceptions. Any error coming from the Wisconsin RPC library results in an exception being thrown (a single type is used).

*[feature was dropped]*
The optional stack trace functionality works by inserting a hook in the low-level API used by `throw` statements. More specifically, it overrides `__cxa_throw` (also from the Itanium ABI) to record a stack trace every time an exception is thrown. When an exception is caught in `invoke`, the stack trace is queried and sent back to the caller.

Since recording a stack trace for every exception adds some overhead, it is made optional using a preprocessor macro. If `NDEBUG` is set when `register.h` is included, no trace will be recorded. (I'm not sure whether this applies at the level of a module, a compilation unit or even the part of it after the `#include`. This would need to be tested.)

### CTP7 warning

*[feature was dropped, irrelevant]*
The version of `libstdcxx` on the CTP7 has a bug that prevents getting a stack trace. It also fools `gdb`. The best option seems to ask Wisconsin for an image based on a more recent Petalinux. We could also provide an updated version of `libstdcxx` to `LD_PRELOAD` when starting `wiscrpcsvc`, or a tiny library that would only patch the buggy function.

The code was tested primarily on x86, with some testing on a more recent ARM Linux.

### Types of changes

## Motivation and Context

We want to catch exceptions thrown by RPC modules, and provide devs with as much debugging information as possible.

## How Has This Been Tested?

All prototyping was done on x86. Testing on ARM (CTP7 and another chip) was performed to check for architecture-dependent issues. This polished version was tested on x86.

The following client was used with an RPC-ified `memory`:
```C++
#include "memory.h"

#include <iomanip>
#include <iostream>

#include "xhal/rpc/call.h"

int main(int argc, char **argv)
{
    try
    {
        wisc::RPCSvc conn;
        conn.connect("localhost");

        conn.load_module("memory", "memory v1.0.1");

        auto mem = xhal::rpc::call<Memory::Read>(conn, 0, 10);
        std::cout << std::hex;
        for (std::uint32_t word : mem) {
            std::cout << " " << word;
        }
        std::cout << std::endl;
    }
    catch(const xhal::rpc::RemoteException &e)
    {
        std::cerr << "Remote call failed: " << e.what() << std::endl;
        if (e.hasTrace()) {
            std::cerr << e.trace() << std::endl;
        }
        return 1;
    }
    catch(const xhal::rpc::MessageException &e)
    {
        std::cerr << "Remote call failed: " << e.what() << std::endl;
        return 1;
    }

    return 0;
}
```

Compilation was tested on `gem904daq01` with the Xilinx SDK 2016.2.

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,28 +3,42 @@ SUBPACKAGES := \
         xhalcore \
         xhalarm
 
-SUBPACKAGES.DEBUG    := $(patsubst %,%.debug,    $(SUBPACKAGES))
-SUBPACKAGES.RPM      := $(patsubst %,%.rpm,      $(SUBPACKAGES))
-SUBPACKAGES.DOC      := $(patsubst %,%.doc,      $(SUBPACKAGES))
-SUBPACKAGES.CLEANRPM := $(patsubst %,%.cleanrpm, $(SUBPACKAGES))
-SUBPACKAGES.CLEANDOC := $(patsubst %,%.cleandoc, $(SUBPACKAGES))
 SUBPACKAGES.CLEAN    := $(patsubst %,%.clean,    $(SUBPACKAGES))
+SUBPACKAGES.DEBUG    := $(patsubst %,%.debug,    $(SUBPACKAGES))
+SUBPACKAGES.INSTALL  := $(patsubst %,%.install,  $(SUBPACKAGES))
+SUBPACKAGES.UNINSTALL:= $(patsubst %,%.uninstall,$(SUBPACKAGES))
+SUBPACKAGES.RPM      := $(patsubst %,%.rpm,      $(SUBPACKAGES))
+SUBPACKAGES.CLEANRPM := $(patsubst %,%.cleanrpm, $(SUBPACKAGES))
+SUBPACKAGES.DOC      := $(patsubst %,%.doc,      $(SUBPACKAGES))
+SUBPACKAGES.CLEANDOC := $(patsubst %,%.cleandoc, $(SUBPACKAGES))
 
-.PHONY: build clean cleanall cleandoc cleanrpm
+.PHONY: $(SUBPACKAGES) \
+	$(SUBPACKAGES.CLEAN) \
+	$(SUBPACKAGES.INSTALL) \
+	$(SUBPACKAGES.UNINSTALL) \
+	$(SUBPACKAGES.RPM) \
+	$(SUBPACKAGES.CLEANRPM) \
+	$(SUBPACKAGES.DOC) \
+	$(SUBPACKAGES.CLEANDOC)
 
+.PHONY: all build clean cleanall cleandoc cleanrpm
 build: $(SUBPACKAGES)
 
-all: $(SUBPACKAGES) $(SUBPACKAGES.RPM) $(SUBPACKAGES.DOC)
+all: $(SUBPACKAGES) $(SUBPACKAGES.DOC)
+
+doc: $(SUBPACKAGES.DOC)
 
 rpm: $(SUBPACKAGES) $(SUBPACKAGES.RPM)
 
-doc: $(SUBPACKAGES.DOC)
+clean: $(SUBPACKAGES.CLEAN)
 
 cleanrpm: $(SUBPACKAGES.CLEANRPM)
 
 cleandoc: $(SUBPACKAGES.CLEANDOC)
 
-clean: $(SUBPACKAGES.CLEAN)
+install: $(SUBPACKAGES) $(SUBPACKAGES.INSTALL)
+
+uninstall: $(SUBPACKAGES.UNINSTALL)
 
 cleanall: clean cleandoc cleanrpm
 
@@ -34,22 +48,30 @@ $(SUBPACKAGES):
 $(SUBPACKAGES.RPM): $(SUBPACKAGES)
 	$(MAKE) -C $(patsubst %.rpm,%, $@) rpm
 
-$(SUBPACKAGES.DOC):
-	$(MAKE) -C $(patsubst %.doc,%, $@) doc
-
-$(SUBPACKAGES.CLEANRPM):
-	$(MAKE) -C $(patsubst %.cleanrpm,%, $@) cleanrpm
+$(SUBPACKAGES.CLEAN):
+	$(MAKE) -C $(patsubst %.clean,%, $@) clean
 
 $(SUBPACKAGES.CLEANDOC):
 	$(MAKE) -C $(patsubst %.cleandoc,%, $@) cleandoc
 
-$(SUBPACKAGES.CLEAN):
-	$(MAKE) -C $(patsubst %.clean,%, $@) clean
+$(SUBPACKAGES.CLEANRPM):
+	$(MAKE) -C $(patsubst %.cleanrpm,%, $@) cleanrpm
 
-.PHONY: $(SUBPACKAGES) $(SUBPACKAGES.INSTALL) $(SUBPACKAGES.CLEAN) $(SUBPACKAGES.DOC) $(SUBPACKAGES.RPM) $(SUBPACKAGES.CLEANRPM) $(SUBPACKAGES.CLEANDOC)
+$(SUBPACKAGES.DOC):
+	$(MAKE) -C $(patsubst %.doc,%, $@) doc
+
+$(SUBPACKAGES.INSTALL): $(SUBPACKAGES)
+	$(MAKE) -C $(patsubst %.install,%, $@) install
+
+$(SUBPACKAGES.UNINSTALL): $(SUBPACKAGES)
+	$(MAKE) -C $(patsubst %.uninstall,%, $@) uninstall
 
 python:
 
 xhalarm:
 
-xhalcore: xhalarm
+xhalcore:
+
+xhalcore.install: xhalarm
+
+xhalcore.rpm: xhalarm
--- a/xhalcore/Makefile
+++ b/xhalcore/Makefile
@@ -9,111 +9,127 @@ PackageDir   := pkg/$(ShortPackage)
 Packager     := Mykhailo Dalchenko
 Arch         := x86_64
 
-XHAL_VER_MAJOR:=$(shell $(BUILD_HOME)/$(Project)/config/tag2rel.sh | awk '{split($$0,a," "); print a[1];}' | awk '{split($$0,b,":"); print b[2];}')
-XHAL_VER_MINOR:=$(shell $(BUILD_HOME)/$(Project)/config/tag2rel.sh | awk '{split($$0,a," "); print a[2];}' | awk '{split($$0,b,":"); print b[2];}')
-XHAL_VER_PATCH:=$(shell $(BUILD_HOME)/$(Project)/config/tag2rel.sh | awk '{split($$0,a," "); print a[3];}' | awk '{split($$0,b,":"); print b[2];}')
+ProjectPath:= $(BUILD_HOME)/$(Project)
+ConfigDir  := $(ProjectPath)/config
 
-INSTALL_PREFIX=/opt/xhal
+include $(ConfigDir)/mfCommonDefs.mk
+include $(ConfigDir)/mfPythonDefs.mk
+include $(ConfigDir)/mfRPMRules.mk
 
-include $(BUILD_HOME)/$(Project)/config/mfCommonDefs.mk
-include $(BUILD_HOME)/$(Project)/config/mfPythonDefs.mk
-include $(BUILD_HOME)/$(Project)/config/mfRPMRules.mk
+PackageSourceDir:=$(PackagePath)/src/common
 
-CCFLAGS=-O0 -g3 -fno-inline -Wall -pthread
-ADDFLAGS=-fPIC -std=c++11 -m64
+XHAL_VER_MAJOR:=$(shell $(ConfigDir)/tag2rel.sh | awk '{split($$0,a," "); print a[1];}' | awk '{split($$0,b,":"); print b[2];}')
+XHAL_VER_MINOR:=$(shell $(ConfigDir)/tag2rel.sh | awk '{split($$0,a," "); print a[2];}' | awk '{split($$0,b,":"); print b[2];}')
+XHAL_VER_PATCH:=$(shell $(ConfigDir)/tag2rel.sh | awk '{split($$0,a," "); print a[3];}' | awk '{split($$0,b,":"); print b[2];}')
 
-IncludeDirs+= /opt/xdaq/include
-IncludeDirs+= $(BUILD_HOME)/$(Project)/$(LongPackage)/include
+CCFLAGS=-fno-inline -Wall -pthread
+ADDFLAGS=-g -fPIC -std=c++11 -m64
+
+IncludeDirs+= $(XDAQ_ROOT)/include
+IncludeDirs+= $(PackageIncludeDir)
 INC=$(IncludeDirs:%=-I%)
 
-Libraries+= -llog4cplus -lxerces-c -lwiscrpcsvc -lstdc++
+Libraries+=-llog4cplus -lxerces-c -lwiscrpcsvc -lstdc++
 
-LibraryDirs+=-L/opt/xdaq/lib
+LibraryDirs+=-L$(XDAQ_ROOT)/lib
+LibraryDirs+=-L$(PackageLibraryDir)
 LibraryDirs+=-L/opt/wiscrpcsvc/lib
 
 LDFLAGS = -shared $(LibraryDirs)
 
-SrcLocation  = src/common
-SRCS_UTILS   = $(shell echo $(SrcLocation)/utils/*.cpp)
-SRCS_XHAL    = $(shell echo $(SrcLocation)/*.cpp)
-SRCS_XHALPY  = $(shell echo $(SrcLocation)/python_wrappers/*.cpp)
-SRCS_RPC_MAN = $(shell echo $(SrcLocation)/rpc_manager/*.cpp)
+SRCS_XHAL   = $(wildcard $(PackageSourceDir)/*.cpp)
+SRCS_UTILS  = $(wildcard $(PackageSourceDir)/utils/*.cpp)
+SRCS_XHALPY = $(wildcard $(PackageSourceDir)/python_wrappers/*.cpp)
+SRCS_RPCMAN = $(wildcard $(PackageSourceDir)/rpc_manager/*.cpp)
+# SRCS_EXES     = $(wildcard $(PackageSourceDir)/*.cxx)
+# SRCS_TEST_EXES= $(wildcard $(PackageTestSourceDir)/*.cxx)
 
-## Place object files in src/linux/$(Arch)
-ObjLocation = src/linux/$(Arch)
-OBJS_UTILS   = $(patsubst $(SrcLocation)/%,$(ObjLocation)/%, $(SRCS_UTILS:.cpp=.o))
-OBJS_XHAL    = $(patsubst $(SrcLocation)/%,$(ObjLocation)/%, $(SRCS_XHAL:.cpp=.o))
-OBJS_XHALPY  = $(patsubst $(SrcLocation)/%,$(ObjLocation)/%, $(SRCS_XHALPY:.cpp=.o))
-OBJS_RPC_MAN = $(patsubst $(SrcLocation)/%,$(ObjLocation)/%, $(SRCS_RPC_MAN:.cpp=.o))
+AUTODEPS_XHAL   = $(patsubst $(PackageSourceDir)/%.cpp,$(PackageObjectDir)/%.d,$(SRCS_XHAL))
+AUTODEPS_UTILS  = $(patsubst $(PackageSourceDir)/%.cpp,$(PackageObjectDir)/%.d,$(SRCS_UTILS))
+AUTODEPS_XHALPY = $(patsubst $(PackageSourceDir)/%.cpp,$(PackageObjectDir)/%.d,$(SRCS_XHALPY))
+AUTODEPS_RPCMAN = $(patsubst $(PackageSourceDir)/%.cpp,$(PackageObjectDir)/%.d,$(SRCS_RPCMAN))
 
-XHALCORE_LIB = $(BUILD_HOME)/$(Project)/$(LongPackage)/lib/libxhal.so
-RPC_MAN_LIB  = $(BUILD_HOME)/$(Project)/$(LongPackage)/lib/librpcman.so
-XHALPY_LIB   = $(BUILD_HOME)/$(Project)/$(LongPackage)/lib/xhalpy.so
+OBJS_XHAL   = $(patsubst %.d,%.o,$(AUTODEPS_XHAL))
+OBJS_UTILS  = $(patsubst %.d,%.o,$(AUTODEPS_UTILS))
+OBJS_XHALPY = $(patsubst %.d,%.o,$(AUTODEPS_XHALPY))
+OBJS_RPCMAN = $(patsubst %.d,%.o,$(AUTODEPS_RPCMAN))
 
-.PHONY: clean xhalcore rpc prerpm
+XHAL_LIB   = $(PackageLibraryDir)/libxhal.so
+XHALPY_LIB = $(PackageLibraryDir)/xhalpy.so
+RPCMAN_LIB = $(PackageLibraryDir)/librpcman.so
+
+TargetLibraries:= xhal xhalpy rpcman
+
+# destination path macro we'll use below
+df = $(PackageObjectDir)/$(*F)
+
+.PHONY: xhal rpc
 
 default: build
 	@echo "Running default target"
 	$(MakeDir) $(PackageDir)
 
-_rpmprep: preprpm
-	@echo "Running _rpmprep target"
-
-preprpm: default
+rpmprep: default
 	@echo "Running preprpm target"
 	$(MakeDir) lib/arm
-	@cp -rfp $(BUILD_HOME)/$(Project)/xhalarm/lib/*.so lib/arm
-	@cp -rfp lib $(PackageDir)
+	$(MakeDir) $(PackageDir)/$(Package)
+	@cp -rfp $(ProjectPath)/xhalarm/lib/*.so lib/arm
+	@cp -rfp lib include src Makefile $(PackageDir)/$(Package)
+	@cp -rfp $(ProjectPath)/config $(PackageDir)
+	$(MakeDir) $(RPMBUILD_DIR)/SOURCES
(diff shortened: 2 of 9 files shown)


xhalcore/include/xhal/rpc/register.h
```diff
@@ -4,13 +4,34 @@
  * a remotely callable RPC method
  *
  * \author Laurent Pétré <lpetre@ulb.ac.be>
+ * \author Louis Moureaux <lmoureau@ulb.ac.be>
  */
 
 #ifndef XHAL_RPC_REGSTER_H
 #define XHAL_RPC_REGSTER_H
```
mexanick: REGSTER --> REGISTER

mexanick (changes requested):

mexanick:
Considering asking UW to bake a new PetaLinux distributive: we can ask, but I'm quite sceptical whether it will ever be done... 
Also, shall we extend the exceptions propagation to include warnings alongside the errors?

jsturdy:
> Considering asking UW to bake a new PetaLinux distributive: we can ask, but I'm quite sceptical whether it will ever be done...

Indeed, but if we outline our strict requirements, we *might* get some support (or we can push to basically be allowed to control our own destiny viz the linux core, and blow a `cc7` image if we so desire (as other groups are contemplating), and we should sell our GE1/1 experience as feedback for our GE2/1 system, where we need a more fluid and symbiotic relationship with this side of things

> Also, shall we extend the exceptions propagation to include warnings alongside the errors?

I am really not in favour of using `exceptions` as anything other than `exceptions`...
Can you give an example of something you would consider a "warning" that you'd like to expose in this way?

jsturdy:
One concern/question I have is why do we need something that seemingly has a lot of added code and overhead, when at a minimum, I think what we need is just something that can:
1) `catch` the exception in the `invoke` method running on the CTP7
1) translate this message into an `error` key
1) `throw` the exception (in the event of the presence of the `error` key) on the receiving end in the `call` method

At this point, the code that executed the call to the remote function simply handles the exception as it normally would.

I'd even be in favour of adding some exception classe(s) here (more than the two you have added) since this will be included by **both** `ctp7_modules` and `cmsgemos`), which can then easily allow the transferring of error cause and intent between contexts.

What is the use case for the complicated design pattern?

jsturdy:
I'll also `rebase` this on top of the recently merged #123, unless @lmoureaux is back from holiday, as I'd have to force push to your `github`

xhalcore/include/xhal/rpc/register.h
lmoureaux: Not my code nor the purpose of this PR.

xhalcore/include/xhal/rpc/call.h
```diff
@@ -30,29 +32,140 @@ namespace xhal { namespace rpc {
...
              typename std::enable_if<std::is_base_of<xhal::rpc::Method, Method>::value, int>::type = 0
             >
+    helper::functor_return_t<Method> call(wisc::RPCSvc &connection, Args&&... args);
+
+    /**
+     * \brief Thrown by \ref call when an exception is thrown on the remote host.
+     */
+    class RemoteException : public std::runtime_error
+    {
+        bool m_hasType = false;
```
jsturdy: One other thing, is there a need for these `bool` types?
Can't they equally well be encapsulated in checking whether the `m_type`/`m_trace` strings are empty?

lmoureaux: "The type info was not sent" is slightly different than "the type info was empty" but we can live without this distinction. Your call.

jsturdy: What is the foreseen (or actual) use case for sending an empty type?

lmoureaux: Only a really, really bad problem (like being out of memory) could ever cause the type or backtrace to not be present. I'll drop the fields.

lpetre-ulb: I agree it would simplify the API if you drop the fields. Even more since `m_hasTrace` is already defined as `m_trace.empty()`. ;-) Out of curiosity, do you think we would even get a RPC reply if a problem that prevents the type to be recovered occurs?

xhalcore/include/xhal/rpc/call.h
```diff
@@ -30,29 +32,133 @@ namespace xhal { namespace rpc {
...
+        friend helper::functor_return_t<Method> call(wisc::RPCSvc &connection,
+                                                     Args&&... args);
+
+        /**
+         * \brief Constructor.
+         * \param response An RPC response to extract error information from.
+         */
+        explicit RemoteException(const wisc::RPCMsg &response) :
+            std::runtime_error(helper::getExceptionMessage(response)),
+            m_type(response.get_key_exists("type") ? response.get_string("type") : ""),
```
lpetre-ulb: Shouldn't the `type` and ` backtrace` RPC keys be prefixed with some templated RPC specific namespace, such as `abiVersion`? In my mind it was one of the purposes of `abiVersion`. Moreover it would help preventing name clashes with any upgrade of `rpcsvc` although extremely unlikely and poorly documented...

xhalcore/include/xhal/rpc/call.h
```diff
@@ -30,29 +32,133 @@ namespace xhal { namespace rpc {
...
+        std::string m_type;
+        std::string m_trace;
+
+        /**
+         * \brief \ref call is the only function that can throw this exception.
+         */
+        template<typename Method,
+             typename... Args,
+             typename std::enable_if<std::is_base_of<xhal::rpc::Method, Method>::value, int>::type
+            >
```
lpetre-ulb: Do we have a style for the multi-line templates? I was aligning the `typename`'s as well as the `<` and `>` (see L.31-34). You use 4 spaces which is one indentation, but I consider this more as alignment than indentation. @jsturdy? @mexanick?

lmoureaux:
Rebased on top of `feature/templated-rpc-methods`; implemented Jared's comment; fixed an issue with exceptions from the standard library.

Will look into build system integration next.

lmoureaux:
Also added an example backtrace to the summary above.

xhalcore/include/xhal/rpc/call.h
```diff
@@ -30,29 +32,133 @@ namespace xhal { namespace rpc {
...
+            // Remote call
+            const wisc::RPCMsg response = connection.call_method(request);
 
-        // Type conversion from args to serializable types
-        // must be performed in the same statement in order to
-        // make use of lifetime extension
-        query << helper::get_forward_as_tuple<Method>()(args...);
+            // Check for errors
+            if (response.get_key_exists("rpcerror")) {
+                throw MessageException(response.get_string("rpcerror"));
```
lpetre-ulb, line 130: Can the `rpcerror` key be considered as a `MessageException`? Looking at your documentation it is indeed related to an error in the underlying Wisconsin library, but if I did not miss anything this (undocumented) case only occurs when the called RPC method does not exist. So only if the templated RPC framework fails or is misused. This comment is nitpicking though, so you can ignore this... or rename the exception.

xhalcore/include/xhal/rpc/call.h
```diff
@@ -30,29 +32,133 @@ namespace xhal { namespace rpc {
...
-        return return_v.get();
+        } catch (const wisc::RPCMsg::BadKeyException &e) {
+            throw MessageException(helper::getExceptionMessage(e));
+        } catch (const wisc::RPCMsg::TypeException &e) {
+            throw MessageException(helper::getExceptionMessage(e));
+        } catch (const wisc::RPCMsg::BufferTooSmallException &e) {
+            throw MessageException(helper::getExceptionMessage(e));
+        } catch (const wisc::RPCMsg::CorruptMessageException &e) {
+            throw MessageException(helper::getExceptionMessage(e));
+        } catch (const wisc::RPCSvc::RPCException &e) {
```
lpetre-ulb, line 151: Since the RPC Wisconsin exceptions do not inherit from a base class, are you sure you have all the exception types which can be thrown?

xhalcore/include/xhal/rpc/exceptions.h
```diff
@@ -0,0 +1,90 @@
...
+
+#include "xhal/rpc/wiscrpcsvc.h"
+
+namespace xhal::rpc::helper {
+
+    /**
+     * \brief Retrieves a user-friendly message for the given exception.
+     */
+    template<typename Exception>
+    std::string getExceptionMessage(const Exception &e) = delete;
```
lpetre-ulb: Do you have specific reasons to use templates+overloads instead of simple overloads? Do you fear an unexpected implicit conversion due to the lack of `explicit` constructors?

xhalcore/include/xhal/rpc/exceptions.h
```diff
@@ -0,0 +1,90 @@
...
+     *           retrieve the type name of the current exception.
+     */
+    void setExceptionType(wisc::RPCMsg *response);
+
+    /**
+     * \brief Fetches an error message from \c response.
+     *
+     * The \c error key must be set and to a string.
+     */
+    std::string getExceptionMessage(const wisc::RPCMsg &response);
```
lpetre-ulb: Why is this function not a template specialization while the others `getExceptionMessage` functions are?

lpetre-ulb (commented):
Quite good, mostly questions about details of the implementation and about the style.

lpetre-ulb:
> > Considering asking UW to bake a new PetaLinux distributive: we can ask, but I'm quite sceptical whether it will ever be done...
> 
> Indeed, but if we outline our strict requirements, we _might_ get some support (or we can push to basically be allowed to control our own destiny viz the linux core, and blow a `cc7` image if we so desire (as other groups are contemplating), and we should sell our GE1/1 experience as feedback for our GE2/1 system, where we need a more fluid and symbiotic relationship with this side of things

As told in the first post, the issue lies in `libstdcxx` and even prevents `gdb` (via `gdbserver`) to get full backtraces. The issue is [documented and fixed upstream](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56846). Our minimal requirement would then be to upgrade GCC from version 4.9.2 to version 4.9.3, even though a more recent version would definitively be better.

About CC7, do you know if an ARM release is provided? If yes, is it fully compatible with the kernel provided by Xilinx. If no, a more recent PetaLinux releases with more mainstream technology (such as overlay filesystems) would help.

> I'd even be in favour of adding some exception classe(s) here (more than the two you have added) since this will be included by **both** `ctp7_modules` and `cmsgemos`), which can then easily allow the transferring of error cause and intent between contexts.

I fear defining more exception types would defeat the abstraction implemented in the templated RPC framework. If they were defined here, it would increase the difficulty to create standalone tools and if they were defined in the `ctp7_modules`, the templated RPC would have to follow all changes in the `ctp7_modules`. The only abstracted way I currently see is to register the exception classes via a template mechanism.

xhalcore/include/xhal/rpc/call.h
jsturdy: Personally, I would also align the `typenames`, but that may just be my aesthetic preference...

jsturdy:
> > > Considering asking UW to bake a new PetaLinux distributive: we can ask, but I'm quite sceptical whether it will ever be done...
> > 
> > 
> > Indeed, but if we outline our strict requirements, we _might_ get some support (or we can push to basically be allowed to control our own destiny viz the linux core, and blow a `cc7` image if we so desire (as other groups are contemplating), and we should sell our GE1/1 experience as feedback for our GE2/1 system, where we need a more fluid and symbiotic relationship with this side of things
> 
> As told in the first post, the issue lies in `libstdcxx` and even prevents `gdb` (via `gdbserver`) to get full backtraces. The issue is [documented and fixed upstream](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56846). Our minimal requirement would then be to upgrade GCC from version 4.9.2 to version 4.9.3, even though a more recent version would definitively be better.

This is probably possible, but the primary issue might be that UW are not really supporting CTP7s, and have moved resources to future boards, so they really only provide some limited support.  Again, it can't hurt to ask, and for something like this, I'd be optimistic (unlike say for the inclusion of a version with `overlayfs`...)

> 
> About CC7, do you know if an ARM release is provided? If yes, is it fully compatible with the kernel provided by Xilinx. If no, a more recent PetaLinux releases with more mainstream technology (such as overlay filesystems) would help.

I only know that there are other groups working on LHC projects who are putting `centos` variants on things like the `Zynq`.  I believe that there is also some work being done to determine if an officially supported CERN release would be feasible.

> 
> > I'd even be in favour of adding some exception classe(s) here (more than the two you have added) since this will be included by **both** `ctp7_modules` and `cmsgemos`), which can then easily allow the transferring of error cause and intent between contexts.
> 
> I fear defining more exception types would defeat the abstraction implemented in the templated RPC framework. If they were defined here, it would increase the difficulty to create standalone tools and if they were defined in the `ctp7_modules`, the templated RPC would have to follow all changes in the `ctp7_modules`. The only abstracted way I currently see is to register the exception classes via a template mechanism.

To be clear, I'm not saying we should have arbitrary exceptions, rather, specific named exceptions inheriting from these two basic types (even on-the-fly exception creation macros, a la `xcept::Exception` from `xdaq`), which is potentially interesting from the handling point of view; however, if **any** remote exception is going to trigger some error condition, this may be a moot point...

jsturdy:
> Rebased on top of `feature/templated-rpc-methods`; implemented Jared's comment; fixed an issue with exceptions from the standard library.
> 
> Will look into build system integration next.

@lmoureaux, you've `rebase`d on top of the wrong branch:
```
* b006432        (louis/feature/templated-rpc-methods-exceptions) Drop boolean fields in RemoteException
* 02836fd        Add code to record the stack trace on throw
* a60eccb        Update \author info
* 82b6b08        Add support for sending back remote stack traces
* 69ce392        Handle RPCMsg and RPCSvc exceptions in call()
* eb5ea95        Handle the error key in call()
* 198386d        Catch exceptions thrown by RPC methods
| * 1c95eb5  (HEAD -> feature/templated-rpc-methods-exceptions) Add code to record the stack trace on throw
| * 4407b02  Update \author info
| * f25ec24  Add support for sending back remote stack traces
| * c690806  Handle RPCMsg and RPCSvc exceptions in call()
| * b190de8  Handle the error key in call()
| * dc54e5d  Catch exceptions thrown by RPC methods
| *   a4d188c        (gemdaq/feature/templated-rpc-methods, feature/templated-rpc-methods) Merge pull request #123 from lpetre-ulb/feature/templated-rpc-methods
| |\
| |/
|/|
* | dba680f  (louis/feature/templated-rpc-methods, laurent/feature/templated-rpc-methods) Relax namespace constraints for custom type serializers
* | 946f5e5  Add support for custom types
* | e259268  Add support for `std::map<std:string, T>`
* | 66fbbbf  Add support of `std::map<std::uint32_t, T>`
* | 2bd6734  Add support for `std::array` of integral types
| |
|/
*   32b6094  Merge pull request #121 from lpetre-ulb/feature/templated-rpc-methods
```

xhalcore/include/xhal/rpc/call.h
lmoureaux: Will update the docs.

xhalcore/include/xhal/rpc/call.h
lmoureaux: Yes

xhalcore/include/xhal/rpc/exceptions.h
lmoureaux: I guess I forgot about overloads when learning about templates...

xhalcore/include/xhal/rpc/exceptions.h
lmoureaux: It has a different purpose. Will rename it.

xhalcore/include/xhal/rpc/call.h
lmoureaux: It looks better than using one indentation. An alternative would be:

```C++
template<
    typename foo,
    typename bar = something_very_long_that_prevents_putting_it_on_a_single_line
>
```

I'll align everything to the opening `<`

lmoureaux:
* Rebased on top of the correct branch
* Addressed all comments. Since I was rebasing anyway, edited the history
* Dropped the backtrace feature as discussed this morning; it will go into a separate PR

Still no build system integration.

xhalcore/include/xhal/rpc/register.h
mexanick: c'mon, this is one-liner fix and certainly a typo introduced before ;)

jsturdy:
> Still no build system integration.

Blocker

lmoureaux:
Added build system integration. Symbols are added to `libxhal.so`. Proof:
```
[lmoureau@gem904daq01 xhal]$ nm -a xhalcore/lib/libxhal.so | c++filt | grep ::getExc
00000000000343fb T xhal::rpc::helper::getExceptionMessage(wisc::RPCMsg::TypeException const&)
00000000000343ce T xhal::rpc::helper::getExceptionMessage(wisc::RPCMsg::BadKeyException const&)
0000000000034462 T xhal::rpc::helper::getExceptionMessage(wisc::RPCMsg::BufferTooSmallException const&)
00000000000344c9 T xhal::rpc::helper::getExceptionMessage(wisc::RPCMsg::CorruptMessageException const&)
00000000000344f6 T xhal::rpc::helper::getExceptionMessage(wisc::RPCSvc::RPCException const&)
0000000000034351 T xhal::rpc::helper::getExceptionMessage(std::exception const&)
[lmoureau@gem904daq01 xhal]$ nm -a xhalarm/lib/libxhal.so | c++filt | grep ::getExc
00024580 T xhal::rpc::helper::getExceptionMessage(wisc::RPCMsg::TypeException const&)
00024548 T xhal::rpc::helper::getExceptionMessage(wisc::RPCMsg::BadKeyException const&)
000245e8 T xhal::rpc::helper::getExceptionMessage(wisc::RPCMsg::BufferTooSmallException const&)
00024650 T xhal::rpc::helper::getExceptionMessage(wisc::RPCMsg::CorruptMessageException const&)
00024688 T xhal::rpc::helper::getExceptionMessage(wisc::RPCSvc::RPCException const&)
000244d0 T xhal::rpc::helper::getExceptionMessage(std::exception const&)
```

xhalcore/include/xhal/rpc/register.h
lmoureaux: Did it now that history editing is (hopefully) over, so conflicts are less annoying.

xhalcore/Makefile
```diff
@@ -9,48 +9,64 @@ PackageDir   := pkg/$(ShortPackage)
...
 LibraryDirs+=-L/opt/wiscrpcsvc/lib
 
 LDFLAGS = -shared $(LibraryDirs)
 
-SrcLocation  = src/common
-SRCS_UTILS   = $(shell echo $(SrcLocation)/utils/*.cpp)
-SRCS_XHAL    = $(shell echo $(SrcLocation)/*.cpp)
-SRCS_XHALPY  = $(shell echo $(SrcLocation)/python_wrappers/*.cpp)
-SRCS_RPC_MAN = $(shell echo $(SrcLocation)/rpc_manager/*.cpp)
+SRCS_XHAL    = $(wildcard $(PackageSourceDir)/*.cpp)
```
jsturdy: Question for @mexanick, what is/was the rationale for separating `*_XHAL` and `*_UTILS`?
In this `Makefile` they all simply feed into the `XHAL_LIB` shared object.

Makefile
```diff
@@ -52,4 +52,6 @@ python:
 
 xhalarm:
 
-xhalcore: xhalarm
+xhalcore:
+
+xhalcore.RPM: xhalarm
```
jsturdy: Should be `xhalcore.rpm`, typo was in my stuff.
I also made a few further optimizations to these as well as the templates in `config`.
I don't think you need to include them in this PR, but I'll work them in during the release prep (they also impact other packages)

jsturdy (changes requested):

Makefile
lmoureaux: I can edit your commit if you want stuff to be fixed. Just push fixes to #129 and ping me.

jsturdy: The `config` submodule was updated to cms-gem-daq-project/gembuild#4, and I'll fix the `xhal` commit with the merged hash of this, this afternoon (currently it's pushed with the hash of the pending PR if you want to test everything except the final submodule hash)

jsturdy: `hotfix/simplify-makefiles` is now updated to the merged hash of the `config` submodule

lmoureaux: You rebased it on top of a WIP commit that I removed yesterday.

jsturdy: fixed

lmoureaux: > fixed

Not really (the same commit message appears twice in history):
```
7eb3190 Clean up Makefiles -- Fixes #128
055b8aa Fix typo in rpc/register.h header guard
f577bf2 Compile rpc/exceptions.cpp into libxhal.so
9d9df61 Clean up Makefiles -- Fixes #128
```
I sorted it out. I took the freedom to drop the changes that added the `rpc` directory, since I prefer having them in a separate commit.

jsturdy: Ah, OK, thanks.
I missed that my rebase, rebased onto yours which had already included the commit

jsturdy (approved):
OK from me, just the final decision on https://github.com/cms-gem-daq-project/xhal/pull/123#pullrequestreview-265454420, which can be made in a separate cleanup PR

mexanick (approved):

PR #131, opened 2019-08-25
LMDB guard for CTP7 modules migration

lmoureaux:
## Description

This PR adds a guard that abstracts away the management of the register address database. The current implementation doesn't try to be smart and does the same as the current `localArgs` initialization.

The new code cannot be compiled into `libxhal.so` for the host PC (unless we want to carry LMDB as a dependency), therefore I put the files in the `xhalarm` package structure. However I can't find a way to include them into the build without duplicating most of the Makefile; @jsturdy maybe you can suggest something? I think the canonical way of doing this is to build  `libxhalcore.so` for two architectures from the same code and Makefile (with CPP, CC, CXX, LD etc passed from a root Makefile using the environment), and to build `libxhalarm.so` for ARM, with only the CTP7-specific code.

### Types of changes

## Motivation and Context

Required to port away from `localArgs`.

## How Has This Been Tested?

Compiles and links on x86 with CMake. More testing is needed on a real CTP7, but build system integration is required before this can happen.

## Checklist:

### DIFF
--- a/xhalcore/include/lmdb++.h
+++ b/xhalcore/include/lmdb++.h
@@ -1,0 +1,1927 @@
+/* This is free and unencumbered software released into the public domain. */
+
+#ifndef LMDBXX_H
+#define LMDBXX_H
+
+/**
+ * <lmdb++.h> - C++11 wrapper for LMDB.
+ *
+ * @author Arto Bendiken <arto@bendiken.net>
+ * @see https://sourceforge.net/projects/lmdbxx/
+ */
+
+#ifndef __cplusplus
+#error "<lmdb++.h> requires a C++ compiler"
+#endif
+
+#if __cplusplus < 201103L
+#if !defined(_MSC_VER) || _MSC_VER < 1900
+#error "<lmdb++.h> requires a C++11 compiler (CXXFLAGS='-std=c++11')"
+#endif // _MSC_VER check
+#endif
+
+////////////////////////////////////////////////////////////////////////////////
+
+#include <lmdb.h>      /* for MDB_*, mdb_*() */
+
+#ifdef LMDBXX_DEBUG
+#include <cassert>     /* for assert() */
+#endif
+#include <cstddef>     /* for std::size_t */
+#include <cstdio>      /* for std::snprintf() */
+#include <cstring>     /* for std::strlen() */
+#include <stdexcept>   /* for std::runtime_error */
+#include <string>      /* for std::string */
+#include <type_traits> /* for std::is_pod<> */
+
+namespace lmdb {
+  using mode = mdb_mode_t;
+}
+
+////////////////////////////////////////////////////////////////////////////////
+/* Error Handling */
+
+namespace lmdb {
+  class error;
+  class logic_error;
+  class fatal_error;
+  class runtime_error;
+  class key_exist_error;
+  class not_found_error;
+  class corrupted_error;
+  class panic_error;
+  class version_mismatch_error;
+  class map_full_error;
+  class bad_dbi_error;
+}
+
+/**
+ * Base class for LMDB exception conditions.
+ *
+ * @see http://symas.com/mdb/doc/group__errors.html
+ */
+class lmdb::error : public std::runtime_error {
+protected:
+  const int _code;
+
+public:
+  /**
+   * Throws an error based on the given LMDB return code.
+   */
+  [[noreturn]] static inline void raise(const char* origin, int rc);
+
+  /**
+   * Constructor.
+   */
+  error(const char* const origin,
+        const int rc) noexcept
+    : runtime_error{origin},
+      _code{rc} {}
+
+  /**
+   * Returns the underlying LMDB error code.
+   */
+  int code() const noexcept {
+    return _code;
+  }
+
+  /**
+   * Returns the origin of the LMDB error.
+   */
+  const char* origin() const noexcept {
+    return runtime_error::what();
+  }
+
+  /**
+   * Returns the underlying LMDB error code.
+   */
+  virtual const char* what() const noexcept {
+    static thread_local char buffer[1024];
+    std::snprintf(buffer, sizeof(buffer),
+      "%s: %s", origin(), ::mdb_strerror(code()));
+    return buffer;
+  }
+};
+
+/**
+ * Base class for logic error conditions.
+ */
+class lmdb::logic_error : public lmdb::error {
+public:
+  using error::error;
+};
+
+/**
+ * Base class for fatal error conditions.
+ */
+class lmdb::fatal_error : public lmdb::error {
+public:
+  using error::error;
+};
+
+/**
+ * Base class for runtime error conditions.
+ */
+class lmdb::runtime_error : public lmdb::error {
+public:
+  using error::error;
+};
+
+/**
+ * Exception class for `MDB_KEYEXIST` errors.
+ *
+ * @see http://symas.com/mdb/doc/group__errors.html#ga05dc5bbcc7da81a7345bd8676e8e0e3b
+ */
+class lmdb::key_exist_error final : public lmdb::runtime_error {
+public:
+  using runtime_error::runtime_error;
+};
+
+/**
+ * Exception class for `MDB_NOTFOUND` errors.
+ *
+ * @see http://symas.com/mdb/doc/group__errors.html#gabeb52e4c4be21b329e31c4add1b71926
+ */
+class lmdb::not_found_error final : public lmdb::runtime_error {
+public:
+  using runtime_error::runtime_error;
+};
+
+/**
+ * Exception class for `MDB_CORRUPTED` errors.
+ *
+ * @see http://symas.com/mdb/doc/group__errors.html#gaf8148bf1b85f58e264e57194bafb03ef
+ */
+class lmdb::corrupted_error final : public lmdb::fatal_error {
+public:
+  using fatal_error::fatal_error;
+};
+
+/**
+ * Exception class for `MDB_PANIC` errors.
+ *
+ * @see http://symas.com/mdb/doc/group__errors.html#gae37b9aedcb3767faba3de8c1cf6d3473
+ */
+class lmdb::panic_error final : public lmdb::fatal_error {
+public:
+  using fatal_error::fatal_error;
+};
+
+/**
+ * Exception class for `MDB_VERSION_MISMATCH` errors.
+ *
+ * @see http://symas.com/mdb/doc/group__errors.html#ga909b2db047fa90fb0d37a78f86a6f99b
+ */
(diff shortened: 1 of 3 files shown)


lpetre-ulb:
> The new code cannot be compiled into `libxhal.so` for the host PC (unless we want to carry LMDB as a dependency), therefore I put the files in the `xhalarm` package structure. However I can't find a way to include them into the build without duplicating most of the Makefile; @jsturdy maybe you can suggest something? I think the canonical way of doing this is to build `libxhalcore.so` for two architectures from the same code and Makefile (with CPP, CC, CXX, LD etc passed from a root Makefile using the environment), and to build `libxhalarm.so` for ARM, with only the CTP7-specific code.

It seems the header file is also inside `xhalarm`. How does the header file gets installed on the development machine and how is it included in the client application (during compilation only since RPC methods can (will?) inherit from `xhal::LMDBGuard`)? It is installed in the same location as the `xhalcore` header files?

To me, what's happening with the `xhal` library is very strange. We are getting two libraries with the same name, but which provide different features and symbols depending on the target. Shouldn't we try to rationalize and create one library per functionality (e.g. `libhal_rpc.so`, 'libxhal_lmdb.so`, `libxhal_utils.so`, ...) and compile the libraries only for the supported targets? Features and symbols would be the same, even though the implementation can differ.

xhalarm/src/common/LMDB.cpp
```diff
@@ -0,0 +1,120 @@
...
+            throw std::runtime_error("Out of LMDB guard handles");
+        } else {
+            GUARD_COUNT++; // Add a reference
+        }
+    }
+
+    LMDBGuard::~LMDBGuard()
+    {
+        GUARD_COUNT--;
+        if (GUARD_COUNT <= 0) {
```
jsturdy: Merge two statements?
```c++
if (--GUARD_COUNT <= 0) {
    // blah
```

lmoureaux: Considered it and deemed the readability drawback more important than concision in this simple function.

jsturdy:
> > The new code cannot be compiled into `libxhal.so` for the host PC (unless we want to carry LMDB as a dependency), therefore I put the files in the `xhalarm` package structure. However I can't find a way to include them into the build without duplicating most of the Makefile; @jsturdy maybe you can suggest something? I think the canonical way of doing this is to build `libxhalcore.so` for two architectures from the same code and Makefile (with CPP, CC, CXX, LD etc passed from a root Makefile using the environment), and to build `libxhalarm.so` for ARM, with only the CTP7-specific code.
> 
> It seems the header file is also inside `xhalarm`. How does the header file gets installed on the development machine and how is it included in the client application (during compilation only since RPC methods can (will?) inherit from `xhal::LMDBGuard`)? It is installed in the same location as the `xhalcore` header files?
> 

Currently, the MO has been that if we're developing applications that depend on, e.g., `xhal`, whether for `arm` or for `x86_64`, we need headers, and if for `arm` the `arm` compatible libraries for anything that needs to be compiled for the `arm` architecture.  So during the RPM packaging stage, the `arm` compatible libs (and eventually `arm` specific headers, though there hasn't yet been such a package) are copied into the `RPMBUILD` tree for packaging, such that they can be installed with the `xhal-devel` package.  We could do the equivalent of putting the `arm` libs into `lib/arm` for the headers with an `include/xhal/arm` or `include/arm/xhal` (probably the second makes the code more adaptable)

What we *could* (*should*?) do, is somehow create our own `PETA_STAGE` images, (or an extension package), which can be updated with the `devel` packages that we're producing, as this also helps to keep the wall between `arm` and `x86_64`

> To me, what's happening with the `xhal` library is very strange. We are getting two libraries with the same name, but which provide different features and symbols depending on the target. Shouldn't we try to rationalize and create one library per functionality (e.g. `libhal_rpc.so`, 'libxhal_lmdb.so`, `libxhal_utils.so`, ...) and compile the libraries only for the supported targets? Features and symbols would be the same, even though the implementation can differ.

Yes, this is probably a better route for this package, since its origins in being PC based, are now migrating to a dual use library, with distinct features on each side of the RPC call.
Let's discuss this in #132

xhalarm/src/common/LMDB.cpp
jsturdy: OK... in any case, prefer pre-inc(dec)rement operations (for consistency), *unless* somehow using the underlying difference.

jsturdy:
Tried an attempt to package as being discussed in #132, but the code doesn't compile with the `Zynq` compiler (`arm-linux-gnueabihf-gcc -dumpversion: 4.9.2`) and the following error message:
```
arm-linux-gnueabihf-gcc -fomit-frame-pointer -pipe -fno-common -fno-builtin -Wall -std=c++14 -march=armv7-a -mfpu=neon -mfloat-abi=hard -mthumb-interwork -mtune=cortex-a9 -DEMBED -Dlinux -D__linux__ -Dunix -fPIC --sysroot=/data/bigdisk/sw/peta-stage -I/data/bigdisk/sw/peta-stage/usr/include -I/data/bigdisk/sw/peta-stage/include -g -std=gnu++14 -I/data/bigdisk/use
rs/sturdy/software/xhal/xhal/include/common -I/data/bigdisk/users/sturdy/software/xhal/xhal/include/arm -c -MT /data/bigdisk/users/sturdy/software/xhal/xhalarm/src/linux/arm/arm/LMDB.o -MMD -MP -MF /data/bigdisk/users/sturdy/software/xhal/xhalarm/src/linux/arm/arm/LMDB.Td -o /data/bigdisk/users/sturdy/software/xhal/xhalarm/src/linux/arm/arm/LMDB.o /data/bigdisk/u
sers/sturdy/software/xhal/xhal/src/arm/LMDB.cpp
In file included from /data/bigdisk/users/sturdy/software/xhal/xhal/src/arm/LMDB.cpp:1:0:
/data/bigdisk/users/sturdy/software/xhal/xhal/include/arm/xhal/LMDB.h:60:30: error: defaulted declaration ‘constexpr xhal::LMDBGuard& xhal::LMDBGuard::operator=(xhal::LMDBGuard&&) const’
         constexpr LMDBGuard &operator=(LMDBGuard &&) = default;
                              ^
/data/bigdisk/users/sturdy/software/xhal/xhal/include/arm/xhal/LMDB.h:60:30: error: does not match expected signature ‘xhal::LMDBGuard& xhal::LMDBGuard::operator=(xhal::LMDBGuard&&)’
/data/bigdisk/users/sturdy/software/xhal/xhal/include/arm/xhal/LMDB.h:60:30: error: explicitly defaulted function ‘constexpr xhal::LMDBGuard& xhal::LMDBGuard::operator=(xhal::LMDBGuard&&) const’ cannot be declared as constexpr because the implicit declaration is not constexpr:
/data/bigdisk/users/sturdy/software/xhal/xhal/src/arm/LMDB.cpp: In constructor ‘xhal::{anonymous}::Singleton::Singleton()’:
/data/bigdisk/users/sturdy/software/xhal/xhal/src/arm/LMDB.cpp:43:23: warning: ‘xhal::{anonymous}::Singleton::rtxn’ will be initialized after [-Wreorder]
             lmdb::txn rtxn; ///< \brief Read-only transaction
                       ^
/data/bigdisk/users/sturdy/software/xhal/xhal/src/arm/LMDB.cpp:42:23: warning:   ‘lmdb::dbi xhal::{anonymous}::Singleton::dbi’ [-Wreorder]
             lmdb::dbi dbi;  ///< \brief Database handle
                       ^
/data/bigdisk/users/sturdy/software/xhal/xhal/src/arm/LMDB.cpp:50:13: warning:   when initialized here [-Wreorder]
             Singleton() : // NOTE: Order is important!
             ^
make[1]: *** [/data/bigdisk/users/sturdy/software/xhal/xhalarm/src/linux/arm/arm/LMDB.o] Error 1
make[1]: Leaving directory `/data/bigdisk/users/sturdy/software/xhal/xhalarm'
make: *** [xhalarm] Error 2
```

Removing the `constexpr` allows the compilation to complete... so I can at least continue to poke around with the packaging, but of course best to let the author correct this as intended :-)

jsturdy:
Also, can't compile for "client" because of lack of `std::make_unique` in `gcc-4.8.5`... [workaround](https://stackoverflow.com/a/17902439) can be made, so continuing testing only the packaging, and can open a separate issue about how to handle this.

lmoureaux:
Should now compile without errors or warnings.

xhalarm/src/common/LMDB.cpp
lmoureaux: I *think* the psychological reason for using `x++` instead of `++x` is to keep statements organized as `[affected-variable] [effect];`. I use `--i` in `for` loops though...

Anyway, fixed.

jsturdy:
@mexanick @lmoureaux, any reason to not merge this (as mentioned in https://github.com/cms-gem-daq-project/xhal/issues/132#issuecomment-528561345?

lmoureaux:
I'm reluctant to merging something that doesn't build out-of-the-box. Maybe I can move it to `xhalcore` for the time being? (Since the repo structure will be overhauled anyway)

jsturdy:
> I'm reluctant to merging something that doesn't build out-of-the-box. Maybe I can move it to `xhalcore` for the time being? (Since the repo structure will be overhauled anyway)

Sounds fine to me (I can confirm that it builds after my moving things around though I haven't tested that the library works...)

lmoureaux:
`clang-tidy` finding:
```
xhal/xhalarm/src/common/LMDB.cpp:32:13: warning: Address of stack memory associated with local variable 'env' returned to caller [clang-analyzer-core.StackAddressEscape]
            return std::move(env);
            ^
xhal/xhalarm/src/common/LMDB.cpp:51:21: note: Calling 'create_env'
                env(create_env()),
                    ^
xhal/xhalarm/src/common/LMDB.cpp:25:17: note: Assuming the condition is false
            if (path == nullptr) {
                ^
xhal/xhalarm/src/common/LMDB.cpp:25:13: note: Taking false branch
            if (path == nullptr) {
            ^
xhal/xhalarm/src/common/LMDB.cpp:32:13: note: Address of stack memory associated with local variable 'env' returned to caller
            return std::move(env);
            ^
xhal/xhalarm/src/common/LMDB.cpp:81:16: warning: function previously declared with an explicit exception specification redeclared with an implicit exception specification [clang-diagnostic-implicit-exception-spec-mismatch]
    LMDBGuard::~LMDBGuard()
               ^
xhal/xhalarm/include/xhal/LMDB.h:66:9: note: previous declaration is here
        ~LMDBGuard() noexcept;
        ^
```

lmoureaux:
Fixed `clang-tidy` warnings, moved to `xhalcore`.

jsturdy (approved):

PR #133, opened 2019-09-10
Packaging updates after templated RPC developments

jsturdy:
## Description
This PR adds changes for `gitlab` CI migration along with restructuring of the `xhalarm`/`xhalcore` into a single `xhal` package that builds libraries for different architectures (currently `arm` and `x86_64`)

Changes in the code are whitespace and formatting
* Updated all `#include` directives based on the new structure
* Added additional namespaces, depending on the subpackage (`common`, `client`, `server`)
* Made uniform 4 spaces indentation in all header files (not yet in `cpp` files though this can be added
* Made uniform the style in `cpp` files of not using `namespace blah {` scoping, in favour of `blah::` scoping
* Made uniform the style for `template` declarations:
  * `template<blah>` goes on it's own line
  * multi-line `template<blah,...>` always ends with `>` and the next line starts the next portion of the statement

### Types of changes
- [x] New feature (non-breaking change which adds functionality)

## Motivation and Context
* The CI changes are introduced to facilitate migration to gitlab and automatic publication of releases.
* The restructuring was discussed in #132 and #131 and is seen as the natural progression of the migration to the templated framework

## How Has This Been Tested?
* No code changes were introduced, only formatting.
* The code compiles.
* All the intermediary CI related commits will be squashed before this is merge-ready, but for now, we can treat the code migration as reviewable.

PR #135, opened 2019-12-21
[refactor] structure refactored for consistent building

jsturdy:
## Description
This PR includes all the structural changes for simplified builds.
A subsequent PR can CI integration

### Types of changes

## Motivation and Context
This was added as needed for 

## How Has This Been Tested?
Compilation and packaging has been well tested
Not yet tested with live system.


xhal/include/xhal/common/rpc/compat.h
```diff
@@ -141,6 +144,9 @@ namespace xhal { namespace rpc { namespace compat {
         return {};
     }
 
-} } }
+      }
+    }
+  }
+}
 
 #endif
```
lmoureaux, line 160: `#endif // XHAL_COMMON_RPC_COMPAT_H`

xhal/src/client/XHALDevice.cpp
```diff
@@ -0,0 +1,185 @@
...
+        break;
+      } else {
+        mask = mask >> 1;
+        result = result >> 1;
+      }
+    }
+    return result;
+  } else {
+    XHAL_ERROR("Register not found in address table!");
+    throw xhal::common::utils::XHALXMLParserException(strcat("XHAL XML exception: can't find node", regName.c_str()));
```
lmoureaux, line 65: This use of `strcat` leads to memory corruption. Use `std::string::operator+` instead.

xhal/src/client/XHALDevice.cpp
```diff
@@ -0,0 +1,185 @@
...
+      	rsp = rpc.call_method(req);
+      } STANDARD_CATCH;
+      if (rsp.get_key_exists("error")) {
+        XHAL_ERROR("RPC response returned error, writeReg failed");
+        throw xhal::common::utils::XHALException("Error during register access");
+      }
+    }
+  } else {
+    XHAL_ERROR("Register not found in address table!");
+    throw xhal::common::utils::XHALXMLParserException(strcat("XHAL XML exception: can't find node", regName.c_str()));
```
lmoureaux, line 143: This use of `strcat` leads to memory corruption. Use `std::string::operator+` instead.

python/reg_interface_gem/Makefile
```diff
@@ -25,27 +25,30 @@ include $(ConfigDir)/mfPythonDefs.mk
...
-include $(ConfigDir)/mfSphinx.mk
-
 REG_INTERFACE_GEM_VER_MAJOR:=$(shell $(ConfigDir)/tag2rel.sh | awk '{split($$0,a," "); print a[1];}' | awk '{split($$0,b,":"); print b[2];}')
 REG_INTERFACE_GEM_VER_MINOR:=$(shell $(ConfigDir)/tag2rel.sh | awk '{split($$0,a," "); print a[2];}' | awk '{split($$0,b,":"); print b[2];}')
 REG_INTERFACE_GEM_VER_PATCH:=$(shell $(ConfigDir)/tag2rel.sh | awk '{split($$0,a," "); print a[3];}' | awk '{split($$0,b,":"); print b[2];}')
 
+include $(ConfigDir)/mfPythonRPM.mk
+include $(ConfigDir)/mfSphinx.mk
+
+PythonSources=$(wildcard core/*.py)
```
lmoureaux, line 35: In modern build systems, `wildcard` is considered harmful because:

1. One doesn't notice when new files are introduced, and may fail to add them to the VCS
2. The reviewer may not notice that a file is missing
3. All tests may pass even with a missing file

So in the worst case a product may be shipped with a missing file. Please refrain from using the `wildcard` function.

lmoureaux (changes requested):
* `make` has a strange line at the end:
    ```
    make[2]: Entering directory `/afs/cern.ch/work/l/lmoureau/GEM/test/xhal/xhal'
    make[2]: Nothing to be done for `build'.
    make[2]: Leaving directory `/afs/cern.ch/work/l/lmoureau/GEM/test/xhal/xhal'
    TO DO
    make[1]: Leaving directory `/afs/cern.ch/work/l/lmoureau/GEM/test/xhal/xhal'
    ```
    In the source `TO DO` is always related to documentation.
* `make rpm` fails with:
    ```
    /afs/cern.ch/work/l/lmoureau/GEM/test/xhal/config/ci/generate_repo.sh centos7 x86_64 /afs/cern.ch/work/l/lmoureau/GEM/test/xhal/python/reg_interface_gem/rpm /afs/cern.ch/work/l/lmoureau/GEM/test/xhal/python/reg_interface_gem/rpm/build xhal
    BRANCH_NAME is feature/refactor-structure
    REL_VERSION is unstable-PKG-1.0
    BUILD_VER is 1.0.0
    BUILD_TAG is -final.dev338
    /afs/cern.ch/work/l/lmoureau/GEM/test/xhal/config/ci/generate_repo.sh: line 43: EOS_SITE_URL: unbound variable
    make[2]: *** [_rpmharvest] Error 1
    make[2]: Leaving directory `/afs/cern.ch/work/l/lmoureau/GEM/test/xhal/python/reg_interface_gem'
    make[1]: *** [reg_interface_gem.rpm] Error 2
    make[1]: Leaving directory `/afs/cern.ch/work/l/lmoureau/GEM/test/xhal/python'
    make: *** [python.rpm] Error 2
    ```
    I want a package, not the full YUM repo.
* Pleaase use an automatic formatter like `clang-format`, this will save us from so many style-related comments in the future (pass everything through the formatter and add a CI test to check that `git diff` is clean after doing it again). If you're not satisfied with the options available in CC7 by default, check with less aged versions from `devtoolset`s.

jsturdy:
>  I want a package, not the full YUM repo.

darn it! i had that bug and had fixed it...

xhal/src/client/XHALDevice.cpp
jsturdy: not sure where these came from, but it wasn't introduced by me...

jsturdy:
>   * Pleaase use an automatic formatter like `clang-format`, this will save us from so many style-related comments in the future (pass everything through the formatter and add a CI test to check that `git diff` is clean after doing it again). If you're not satisfied with the options available in CC7 by default, check with less aged versions from `devtoolset`s.

Agree (:-)), but wanted to keep that change separate from this refactor (though, I guess no better time than during a refactor, that is targeting uniformity across multiple projects)

I think I prefer to do all in one subsequent PR that adds 
* `clang-format`/`clang-tidy`
* `black`/`flake8` (for `python`)
* format/rule definitions
* formatted sources

xhal/src/client/XHALDevice.cpp
lmoureaux: I can let it go and open an issue instead

python/reg_interface_gem/Makefile
jsturdy: >   1. One doesn't notice when new files are introduced, and may fail to add them to the VCS
>   2. The reviewer may not notice that a file is missing

I don't really see this as an issue, as any sensible CI build check will catch the missing files, even if the author didn't; 3 I agree is more problematic

lmoureaux: Seems like I forgot the most important:

4. It's impossible to know when a file is removed, and therefore nothing will be recompiled.

lmoureaux:
> formatting

Ok for doing it in another PR.

jsturdy:
> > I want a package, not the full YUM repo.
> 
> darn it! i had that bug and had fixed it...

garrgh, **had** fixed it, and the submodule in the legacy PR *does* include the fix, but the submodule pointer in this branch wasn't updated... go figure

python/reg_interface_gem/Makefile
jsturdy: > Seems like I forgot the most important:
> 
>   4. It's impossible to know when a file is removed, and therefore nothing will be recompiled.

Again, I see this as (almost) the same class as 1 and 2, caught by the CI

jsturdy:
Should be superseded by #133 I think?
At least #133 contains all these commits and a few more (well fewer commits because of some reshuffling and squashing)

lpetre-ulb:
> Should be superseded by #133 I think?

Indeed!

jsturdy:
Superseded by #133

PR #136, opened 2019-12-21
[refactor] refactor legacy codebase

jsturdy:
## Description
applies the same refactor as #135 on the `legacyPreBoost` branch

### Types of changes
Structural change, should have no impact on actual behaviour.
Building may be different

## Motivation and Context
Allow for simpler packaging and testing with locally modified versions of the codebase (and for commonality with the future structure)
Also will make the production of releases standardized across GEM DAQ repos.

## How Has This Been Tested?
Compilation and packaging have been tested extensively


xhal/xhal.mk
```diff
@@ -0,0 +1,250 @@
...
+	$(RM) $(OBJS_XHAL) $(OBJS_RPC_MAN)
+	$(RM) $(PackageLibraryDir)
+	$(RM) $(PackageExecDir)
+	$(RM) $(PackagePath)/$(PackageDir)
+
+cleanall:
+	$(RM) $(PackageObjectDir)
+	$(RM) $(PackagePath)
+
+# default:
```
mexanick, line 211: I guess everything below this line has to be removed

xhalarm/scripts/postinstall.sh
```diff
@@ -1,2 +0,0 @@
-install_path=__prefix__
```
mexanick, line 1: This script would link the newly installed libraries to the place where rpcsvc will look for them. How this will be managed in future layout?

xhal/xhal.mk
jsturdy: indeed, don't know how it snuck past...

xhalarm/scripts/postinstall.sh
jsturdy: Can you remind me specifically what this action was doing?
I have in my mind that the RPM itself will now correctly include symlinks (though it is not **yet** doing so)

mexanick: I think this part was indeed somewhat historical/dev thing. Historically, the rpcsvc will look for libs under `/mnt/persistent/gemdaq/lib/` however, the rpm default install path is `/opt/<package_name>`. So in order to preserve full compatibility (and eventual ease of debugging/rollbacks) the postinstall script has been introduced, which would create symlinks from `/mnt/persistent/gemdaq/lib/` to `/opt/<package_name>` so that subsequent restart of the `rpcsvc` would be sufficient for picking up new libs, but backward compatibility with setup scripts is preserved. I guess now we can change this and alter the setup scripts and default env variables on the CTP7 (or its emulator), then such postinstall script will not be required anymore.

jsturdy: OK, so now, for non `x86_64` libs, they will be packaged in `/mnt/persistent/<package-name>/lib`, but this can very easily be changed/symlinked to `/mnt/persistent/gemdaq/lib`.
With the exception of `ctp7_modules`, we could apply this by default, since they live in a special place already

lpetre-ulb:
Does this PR still make sense since the legacy releases are going to be produced as-is?

jsturdy:
Probably not

PR #144, opened 2020-03-08
Implement new (de)serializers for the RPC calls

### LINKED ISSUE #138: [feature request] Finalize type (de)serializers
## Brief summary of issue
As discussed in cms-gem-daq-project/ctp7_modules#161, there are several missing type (de)serializers.

This is one of the remaining pieces for a released `xhal` in the land of the templated framework.

### Types of issue
- [x] Feature request (request for change which adds functionality)

## Expected Behavior
(de)serializers should be provided for all "common" types

Missing are:
* `std::vector<T>`
* `bool`
* `float`
* `uintX_t` for certain `X` (8 and 16 at least)
* possible `std::unordered_map<K,T>`

lpetre-ulb:
## Description

This PR mainly introduces default new type (de)serializers for the templated RPC calls. In order to fix #138, the following additional types are now supported:

* `bool`;
* `std::uint8_t`;
* `std::uin16_t`;
* `float`;
* `std::vector<T>` where `T` is a (de)serializable type.

To get the best support, split (de)serializers can now be created with the names `save` and `load`. Please refer to the documentation and commit message for more information.

Compilation has also been fixed. Previously, the xDAQ headers for the DAQ machine were searched for during compilation for the CTP7, leading to un-resolvable symbols at run-time.

### Types of changes

## Motivation and Context

See https://github.com/cms-gem-daq-project/xhal/issues/138; used in https://github.com/cms-gem-daq-project/ctp7_modules/pull/169.

## How Has This Been Tested?

Works with https://github.com/cms-gem-daq-project/ctp7_modules/pull/169. More thorough tests should be perform in the future.

## Checklist:

### DIFF
--- a/xhal/include/xhal/common/rpc/common.h
+++ b/xhal/include/xhal/common/rpc/common.h
@@ -83,9 +83,30 @@ namespace xhal {
         wisc::RPCMsg *m_wiscMsg;
 
         /**
-         * @brief Serializes custom types if possible or else supresses implicit type conversions
+         * @brief Adds @c t to the message via a user provided serializer
+         */
+        template<typename T,
+                 typename std::enable_if<
+                   helper::is_serialize_present<MessageSerializer, T>::value, int>::type = 0>
+        inline void saveImpl(const T &t) {
+          // This const_cast is safe when the API is used as intented
+          // More precisely when the object t is modified only with the operator&
+          serialize(*this, const_cast<T &>(t));
+        }
+
+        /**
+         * @brief Adds @c t to the message via a split user provided serializer
+         */
+        template<typename T,
+                 typename std::enable_if<
+                   helper::is_save_present<MessageSerializer, T>::value, int>::type = 0>
+        inline void saveImpl(const T &t) {
+          save(*this, t);
+        }
+
+        /**
+         * @brief Supresses implicit type conversions and warns the user about the error
          *
-         * Every type not defined hereunder is taken care of by this templated function.
          * The function serves two purposes:
          *
          * 1. It delegates the serialization to a well-known function.
@@ -93,11 +114,13 @@ namespace xhal {
          *    remembering the developer that she/he can transmit defined types over the
          *    network.
          */
-        template<typename T>
-        inline void saveImpl(const T &t)
-          // This const_cast is safe when the API is used as intented
-          // More precisely when the object t is modified only with the operator&
-          serialize(*this, const_cast<T &>(t));
+        template<typename T,
+                 typename std::enable_if<
+                   !helper::is_serialize_present<MessageSerializer, T>::value &&
+                   !helper::is_save_present     <MessageSerializer, T>::value, int>::type = 0>
+        inline void saveImpl(const T&) {
+          static_assert(!std::is_same<T,T>::value,
+                        "Type serialization not natively supported. Writing custom serializer is required.");
         }
 
         /**
@@ -265,19 +288,41 @@ namespace xhal {
         const wisc::RPCMsg *m_wiscMsg;
 
         /**
-         * @brief Deserializes custom types if possible or else supresses implicit type conversion
+         * @brief Retrives @c t from the message via a user provided deserializer
+         */
+        template<typename T,
+                 typename std::enable_if<
+                   helper::is_serialize_present<MessageDeserializer, T>::value, int>::type = 0>
+        inline void loadImpl(T &t) {
+          serialize(*this, t);
+        }
+
+        /**
+         * @brief Retrives @c t from the message via a split user provided deserializer
+         */
+        template<typename T,
+                 typename std::enable_if<
+                   helper::is_load_present<MessageDeserializer, T>::value, int>::type = 0>
+        inline void loadImpl(T &t) {
+          load(*this, t);
+        }
+
+        /**
+         * @brief Supresses implicit type conversions and warns the user about the error
          *
          * Every type not defined hereunder is taken care of by this templated function.
-         * The function serves two purposes:
          *
          * 1. It delegates the deserialization to a well-known function.
          * 2. It aims at enforcing maximum type compatibility with the UW RPC API by
          *    reminding the developer that she/he can transmit defined types over the
          *    network.
          */
-        template<typename T>
-        inline void loadImpl(T &t)
-          serialize(*this, t);
+        template<typename T,
+                 typename std::enable_if<
+                   !helper::is_serialize_present<MessageDeserializer, T>::value &&
+                   !helper::is_load_present<MessageDeserializer, T>::value, int>::type = 0>
+        inline void loadImpl(const T&) {
+          static_assert(!std::is_same<T,T>::value, "Type not natively supported. Writing custom serializer is required.");
         }
 
         /**
@@ -425,8 +470,24 @@ namespace xhal {
        * @brief Provides a default (de)serialiazer in case the intrusive method is used
        */
       template<typename Message, typename T>
-      inline void serialize(Message &msg, T &t) {
+      inline auto serialize(Message &msg, T &t) -> decltype(t.serialize(msg)) {
         t.serialize(msg);
+      }
+
+      /**
+       * @brief Provides a default split serialiazer in case the intrusive method is used
+       */
+      template<typename Message, typename T>
+      inline auto save(Message &msg, const T &t) -> decltype(t.save(msg)) {
+        t.save(msg);
+      }
+
+      /**
+       * @brief Provides a default split deserialiazer in case the intrusive method is used
+       */
+      template<typename Message, typename T>
+      inline auto load(Message &msg, T &t) -> decltype(t.load(msg)) {
+        t.load(msg);
       }
 
       /**
@@ -450,6 +511,15 @@ namespace xhal {
        *     // which takes a message as parameter (i.e. the serializer or deserializer)
        *     template<class Message> inline void serialize(Message & msg) {
        *         msg & x & y;
+       *     }
+       *
+       *     // The (de)serializers can also be implemented as two split functions
+       *     // which also take a message as parameter (i.e. the serializer or deserializer)
+       *     template<class Message> inline void save(Message & msg) {
+       *         msg << x << y;
+       *     }
+       *     template<class Message> inline void load(Message & msg) {
+       *         msg >> x >> y;
        *     }
        * };
        *
--- a/xhal/include/xhal/common/rpc/common.h
+++ b/xhal/include/xhal/common/rpc/common.h
@@ -94,7 +94,7 @@ namespace xhal {
          *    network.
          */
         template<typename T>
-        inline void save(const T &t) {
+        inline void saveImpl(const T &t)
           // This const_cast is safe when the API is used as intented
           // More precisely when the object t is modified only with the operator&
           serialize(*this, const_cast<T &>(t));
@@ -103,28 +103,28 @@ namespace xhal {
         /**
          * @brief Adds a @c std::uint32_t to the message
          */
-        inline void save(const std::uint32_t value) {
+        inline void saveImpl(const std::uint32_t value) {
           m_wiscMsg->set_word(std::to_string(dispenseKey()), value);
         }
 
         /**
          * @brief Adds a @c std::vector<std::uint32_t> to the message
          */
-        inline void save(const std::vector<std::uint32_t> &value) {
+        inline void saveImpl(const std::vector<std::uint32_t> &value) {
           m_wiscMsg->set_word_array(std::to_string(dispenseKey()), value);
         }
 
         /**
          * @brief Adds a @c std::string to the message
          */
-        inline void save(const std::string &value) {
+        inline void saveImpl(const std::string &value) {
           m_wiscMsg->set_string(std::to_string(dispenseKey()), value);
         }
 
         /**
          * @brief Adds a @c std::vector<std::string> to the message
          */
-        inline void save(const std::vector<std::string> &value) {
+        inline void saveImpl(const std::vector<std::string> &value) {
           m_wiscMsg->set_string_array(std::to_string(dispenseKey()), value);
         }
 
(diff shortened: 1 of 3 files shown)


xhal/include/xhal/common/rpc/common.h
```diff
@@ -475,6 +546,84 @@ namespace xhal {
           msg & elem;
         }
       }
+
+      /**
+       * @brief (De)erializer for @c std::vector<T> where @c T is a (de)serializable type
```
mexanick: `(De)erializer`->`(De)serializer`

xhal/include/xhal/common/rpc/common.h
```diff
@@ -475,6 +546,84 @@ namespace xhal {
...
+      }
+      template<typename Message>
+      inline void load(Message &msg, std::uint16_t &value) {
+        std::uint32_t tmp;
+        msg >> tmp;
+        value = tmp;
+      }
+
+      /**
+       * @brief (De)serializer for the @float type
```
mexanick, line 610: What are the use cases for floats? This is a bit dangerous and may cause issues in future... I don't see many applications besides ADC conversions which IMHO should not be done on the AMC card (or, if a `float` is really needed we can use strings and convert through them...) I would remove the `float` serialization

mexanick (dismissed):
Cursory approval with few minor comments

xhal/include/xhal/common/rpc/common.h
lpetre-ulb: Fixed !

xhal/include/xhal/common/rpc/common.h
lpetre-ulb: :+1: It was part of the initial request, but I'm not really convinced about the use case. The serializer is required to compile the current version of the RPC module, but IMHO, the RPC module interface should change.
If there is a truly good use, I would agree to convert to strings instead (hidden in the serializer). But, for now, everything compiles as it should. ;-)

xhal/include/xhal/common/rpc/helper.h
```diff
@@ -127,6 +127,34 @@ namespace xhal {
         template<typename T>
         using is_tuple = is_tuple_impl<typename std::decay<T>::type>;
 
+        /**
+         * @brief Define the @c is_func_present class which allows to check the existence of @func
+         *
+         * The @c value member is available to retrieve the lookup result.
+         */
+        #define FUNCTION_PRESENT(func)                                                   \
+        template <typename T, typename A>                                                \
```
lmoureaux: Would it be possible to use less cryptic names for the template parameters? I would also appreciate an example in the Doxygen comment for the macro.

xhal/include/xhal/common/rpc/helper.h
```diff
@@ -127,6 +127,34 @@ namespace xhal {
...
+        class is_##func##_present                                                        \
+        {                                                                                \
+          struct no {};                                                                  \
+                                                                                         \
+          template <typename TT, typename AA>                                            \
+          static decltype(func(std::declval<TT&>(), std::declval<AA&>())) test(int);     \
```
lmoureaux: Please document.

xhal/include/xhal/common/rpc/helper.h
```diff
@@ -127,6 +127,34 @@ namespace xhal {
...
+                                                                                         \
+          template <typename TT, typename AA>                                            \
+          static no test(...);                                                           \
```
lmoureaux: Please document.

xhal/include/xhal/common/rpc/common.h
lmoureaux: Maybe add a `[[deprecated]]` flag right away? I agree it's a bit weird to add deprecated functions, but it would signal that passing `float`s around is not a good idea.

xhal/include/xhal/common/rpc/common.h
```diff
@@ -475,6 +546,84 @@ namespace xhal {
...
+       *
+       * Works only if @sizeof(float) is 4 and the representation 
+       * is the same on the client and the server.
+       */
+      template<typename Message>
+      inline void save(Message &msg, const float &value) {
+        std::uint32_t tmp;
+        std::memcpy(&tmp, &value, sizeof(std::uint32_t));
```
lmoureaux, line 622: Isn't `reinterpret_cast` more appropriate here? ie `msg << *reinterpret_cast<std::uint32_t*>(&value)`

xhal/include/xhal/common/rpc/common.h
```diff
@@ -475,6 +546,84 @@ namespace xhal {
...
+        msg << tmp;
+      }
+      template<typename Message>
+      inline void load(Message &msg, float &value) {
+        std::uint32_t tmp;
+        msg >> tmp;
+        std::memcpy(&value, &tmp, sizeof(float));
```
lmoureaux, line 633: `reinterpret_cast`?

xhal/include/xhal/common/rpc/common.h
```diff
@@ -475,6 +546,84 @@ namespace xhal {
           msg & elem;
         }
       }
+
+      /**
+       * @brief (De)serializer for @c std::vector<T> where @c T is a (de)serializable type
+       */
+      template<typename Message, typename T>
+      inline void serialize(Message &msg, std::vector<T> &value) {
+        // 1. Store or retrieve the vector lenth
```
lmoureaux: `length`

xhal/include/xhal/common/rpc/common.h
```diff
@@ -475,6 +546,84 @@ namespace xhal {
...
+        std::uint32_t length = value.size(); // no-op during deserialization
+        msg & length;
+        value.resize(length); // no-op during serialization
```
lmoureaux, line 559: Note that this requires the value type to be *DefaultInsertable*. I guess that we won't try to serialize anything that fancy.

lmoureaux (changes requested):
In addition to some nitpicking, I'd like to see a clear test plan for `float` since it's not covered by `getVFAT3ChipIDs`.

xhal/include/xhal/common/rpc/common.h
lpetre-ulb: There _may_ be some good use cases, so let's wait until it is completely settled.

xhal/include/xhal/common/rpc/common.h
lpetre-ulb: AFAIK, `reinterpret_cast` is subject to strict type aliasing, while `memcpy` is not. Since we compile with `-O2` or `-O3`, I preferred to use the safest method. Note that compiler will most likely completely optimize the call.

xhal/include/xhal/common/rpc/helper.h
lpetre-ulb: The single letter template parameters name are rather usual naming convention and are used throughout this code base. :+1: for the usage documentation in Doxygen.

xhal/include/xhal/common/rpc/helper.h
lpetre-ulb: How can this be documented? It is a **private** template function declaration which returns the `no` type; there is not definition. Actually, this is very classical C++ template meta-programming code.

lpetre-ulb:
> In addition to some nitpicking, I'd like to see a clear test plan for `float` since it's not covered by `getVFAT3ChipIDs`.

I would have liked to write a test suite, but I have no idea how to build it... If one write the following RPC method:

``` c++
void vfat3::logFloats::operator()(const std::vector<float> &value) const
{
  for (const auto& i : value)
    LOG4CPLUS_WARN(logger, i);
}
```

and trivially calls it within the templated RPC framework: `xhal::common::rpc::call<vfat3::logFloats>(conn, std::vector<float>{3.14, 42.42});`, the log output fully matches the expectations:

```
Mar 11 08:54:51 CTP7 user.warn rpcsvc: eagle48[9449] - void vfat3::logFloats::operator()(const std::vector<float>&) const - 3.14
Mar 11 08:54:51 CTP7 user.warn rpcsvc: eagle48[9449] - void vfat3::logFloats::operator()(const std::vector<float>&) const - 42.42
```

(Notice that it also test `std::vector<T>`.)

xhal/include/xhal/common/rpc/common.h
lpetre-ulb: Good point! But I wouldn't change the (de)serializer for the reason you describe. If someone runs into the issue at some point, the (de)serializer can be improved to handle this case.

xhal/include/xhal/common/rpc/common.h
jsturdy: The historical use case for floating point values has been cases where a register is read many times and an average is computed (whether calculating a rate or mean value or other somesuch) and offloading this to the card makes sense.
One could, of course, return the necessary pieces to calculate the actual desired value elsewhere, but the semantics of that feel a bit weird.
Whether there is still a need or desire for such functionality could be reevaluated...

> If there is a truly good use, I would agree to convert to strings instead (hidden in the serializer). But, for now, everything compiles as it should. ;-)

If you already believe that to be the superior way, why not just implement it as such now?  Or is that implementation significantly more work up front?

jsturdy (commented):
Looks fine modulo comment on `float` and @lmoureaux's docs requests

xhal/include/xhal/common/rpc/common.h
lmoureaux: ok

xhal/include/xhal/common/rpc/common.h
lpetre-ulb: > > If there is a truly good use, I would agree to convert to strings instead (hidden in the serializer). But, for now, everything compiles as it should. ;-)
> 
> If you already believe that to be the superior way, why not just implement it as such now? Or is that implementation significantly more work up front?

As long as the `float` representation is the same on all the machines (and this is IEEE 754 in the extreme majority of the cases), the current implementation is fine and efficient. It would nevertheless be better to have tests about that (and also about all the other (de)serializers).

lmoureaux (dismissed):
I'm satisfied with the testing performed for `float` (and that was the main point). Nitpicking comments can be addressed later.

xhal/include/xhal/common/rpc/common.h
lmoureaux: > It would nevertheless be better to have tests about that

Does testing for `__STDC_IEC_559__` sound like a good compromise?

lmoureaux: Or even better [`std::numeric_limits<float>::is_iec559`](https://en.cppreference.com/w/cpp/types/numeric_limits/is_iec559)

lpetre-ulb: Using `std::numeric_limits<float>::is_iec559` to check the floating point numbers representation (not yet pushed). The use cases from @jsturdy are convincing, so that the feature is not going to be deprecated as soon as it exists. If the functionality is not used (or abused), it can be re-evaluated later on.

lpetre-ulb:
All comments should have been addressed.

xhal/include/xhal/common/rpc/common.h
```diff
@@ -425,10 +472,26 @@ namespace xhal {
        * @brief Provides a default (de)serialiazer in case the intrusive method is used
        */
       template<typename Message, typename T>
-      inline void serialize(Message &msg, T &t) {
+      inline auto serialize(Message &msg, T &t) -> decltype(t.serialize(msg)) {
         t.serialize(msg);
```
lmoureaux, line 476: I guess you're missing a `return` here?

xhal/include/xhal/common/rpc/common.h
```diff
@@ -425,10 +472,26 @@ namespace xhal {
...
       }
 
+      /**
+       * @brief Provides a default split serialiazer in case the intrusive method is used
+       */
+      template<typename Message, typename T>
+      inline auto save(Message &msg, const T &t) -> decltype(t.save(msg)) {
+        t.save(msg);
```
lmoureaux, line 484: Same (`return`)

xhal/include/xhal/common/rpc/common.h
```diff
@@ -425,10 +472,26 @@ namespace xhal {
...
+      }
+
+      /**
+       * @brief Provides a default split deserialiazer in case the intrusive method is used
+       */
+      template<typename Message, typename T>
+      inline auto load(Message &msg, T &t) -> decltype(t.load(msg)) {
+        t.load(msg);
```
lmoureaux, line 492: Same (`return`)

xhal/include/xhal/common/rpc/common.h
```diff
@@ -451,6 +514,15 @@ namespace xhal {
        *     template<class Message> inline void serialize(Message & msg) {
        *         msg & x & y;
        *     }
+       *
+       *     // The (de)serializers can also be implemented as two splt functions
```
lmoureaux: `split` :P

lmoureaux (changes requested):
Missing `return` statements in a few functions returning non-void if (as I understand it) the custom serialization functions do so.

xhal/include/xhal/common/rpc/common.h
lpetre-ulb: (De)serialization functions must return `void`, the `auto-> decltype` is only used to trigger SFINAE.

xhal/include/xhal/common/rpc/common.h
lpetre-ulb: Come on... but fixed!

lmoureaux (approved):

xhal/include/xhal/common/rpc/common.h
mexanick: Time to enable spellchecker in your IDE? ;)

mexanick (approved):

xhal/include/xhal/common/rpc/common.h
lpetre-ulb: Time to copy my text editor configuration to remote machines. ;)

PR #147, opened 2020-03-24
Fix RPC calls with non-reference type parameters

lpetre-ulb:
## Description

[Bug reported by @mexanick]

Before this fix, it was impossible to use the RPC method calls with lvalues if the RPC method parameter type was not a lvalue reference, e.g.

``` bash
/home/mdalchen/cms-gem-daq-project/cmsgemos/gemhardware/devices/src/common/amc/HwGenericAMC.cc:270:120:   required from here
/opt/xhal/include/xhal/common/rpc/call.h:124:58: error: no match for call to ‘(xhal::common::rpc::helper::functor_traits<R (Obj::*)(Args ...) const>::forward_as_tuple() [with Obj = amc::daq::configureDAQModule; R = void; Args = {bool, unsigned int, bool, bool, bool}]::<lambda(bool&&, unsigned int&&, bool&&, bool&&, bool&&)>) (bool&, const unsigned int&, bool&, bool&, bool&)’
           query << helper::get_forward_as_tuple<Method>()(args...);
                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~
/opt/xhal/include/xhal/common/rpc/call.h:124:58: note: candidate: ‘std::tuple<bool&, unsigned int&, bool&, bool&, bool&> (*)(bool&&, unsigned int&&, bool&&, bool&&, bool&&)’ <conversion>
/opt/xhal/include/xhal/common/rpc/call.h:124:58: note:   conversion of argument 6 would be ill-formed:
/opt/xhal/include/xhal/common/rpc/call.h:124:58: error: cannot bind rvalue reference of type ‘bool&&’ to lvalue of type ‘bool’
```

As described in the commit message, the following change has been made to fix the issue. The RPC method calls should now be as close as possible from a real local function call.

> The helper::functor_traits::foward_as_tuple function returns a lambda
function taking the Args&&... argument types (where Args... are the RPC
method argument types). Contrarily to what was expected, the perfect
forwarding template argument syntax only works for deduced arguments,
the compiler needs to know whether the variable is a lvalue or rvalue.
> 
> Since the lambda argument types are not deduced, the following happened:
> 
> * When a reference type was used for the RPC parameter, & collapsed with
  && into &, a lvalue reference.
> * When a non-reference type was used for the RPC parameter, the
  "collapsed" was directly &&, a rvalue reference.
> 
> The solution consists in making sure that all values are captured by
constant lvalue reference in the lambda parameter. The compiler should be
able to optimize the binary as well as before the fix.

### Types of changes

## Motivation and Context

Widely use the templated RPC calls.

## How Has This Been Tested?

Can successfully use this (slightly modified) RPC method:

``` c++
struct getVFAT3ChipIDs : public xhal::common::rpc::Method
{
    std::vector<uint32_t> operator()(const uint16_t ohN, const uint32_t vfatMask=0xff000000, bool rawID=false) const;
};
```

## Checklist:

### DIFF
--- a/xhal/include/xhal/common/rpc/helper.h
+++ b/xhal/include/xhal/common/rpc/helper.h
@@ -47,7 +47,7 @@ namespace xhal {
           using decay_args_type = std::tuple<typename std::decay<Args>::type...>;
 
           static constexpr inline auto forward_as_tuple() {
-            return [](Args&&... args){ return std::forward_as_tuple(args...); };
+            return [](const Args&... args){ return std::forward_as_tuple(args...); };
           }
         };
 


lmoureaux (approved):
Does the extra `const` imply that serializers need be `const`? Or was it already the case?

lpetre-ulb:
> Does the extra `const` imply that serializers need be `const`? Or was it already the case?

Indeed, the arguments to `rpc::call` will be internally passed as `const`. That implies that the serializers must be `const`, as it was already the case (`operator&` and `operator<<` take a `const` parameter).

jsturdy:
> [Bug reported by @mexanick]

Too bad it's not in the issue tracker...

> ## How Has This Been Tested?
> 
> Can successfully use this (slightly modified) RPC method:
> 
> ```c++
> struct getVFAT3ChipIDs : public xhal::common::rpc::Method
> {
>     std::vector<uint32_t> operator()(const uint16_t ohN, const uint32_t vfatMask=0xff000000, bool rawID=false) const;
> };
> ```

Same comment as in cms-gem-daq-project/ctp7_modules#173

mexanick (approved):

lpetre-ulb:
>> Can successfully use this (slightly modified) RPC method:
>> ```c++
>> struct getVFAT3ChipIDs : public xhal::common::rpc::Method
>> {
>>     std::vector<uint32_t> operator()(const uint16_t ohN, const uint32_t vfatMask=0xff000000, bool rawID=false) const;
>> };
>> ```
> 
> Same comment as in [cms-gem-daq-project/ctp7_modules#173](https://github.com/cms-gem-daq-project/ctp7_modules/issues/173)

Right, but in this case, it was only for the test, not for a code-base wide refactoring. ;)

